PR #65, opened 2015-01-06
Add custom user agent to tracker.

tarbrain:
Hello @mattiaslevin,

we have added a feature to make the tracker use a custom user agent when sending tracking data to the backend.

The feature basically consists of an `NSString` property on the `PiwikTracker` class. When set the value is forwarded to the dispatcher instance through the `PiwikDispatcher` protocol and then added to the HTTP header of the HTTP requests.

Best regards
Julian Krumow

### DIFF
--- a/PiwikTracker+AFNetworking2/PiwikAFNetworking2Dispatcher.m
+++ b/PiwikTracker+AFNetworking2/PiwikAFNetworking2Dispatcher.m
@@ -9,7 +9,11 @@
 #import "PiwikAFNetworking2Dispatcher.h"
 #import "AFNetworking.h"
 
+@interface PiwikAFNetworking2Dispatcher ()
 
+@property (nonatomic, strong) NSString *customUserAgent;
+
+@end
 
 @implementation PiwikAFNetworking2Dispatcher
 
@@ -22,6 +26,10 @@
   return self;
 }
 
+- (void)setCustomUserAgent:(NSString *)customUserAgent
+{
+    _customUserAgent = customUserAgent;
+}
 
 - (void)sendSingleEventWithParameters:(NSDictionary*)parameters
                               success:(void (^)())successBlock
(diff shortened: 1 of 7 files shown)


tarbrain:
A custom user agent could provide more detailed information about the users setup.
Especially when using this library:

https://github.com/piwik/device-detector

Basically only the AFNetworkingDispatcher constructs a custom user agent and it only sends the basic device model name like `iPhone` or `iPodTouch`. A more precise value like `iPhone7,2` would be recognized as `iPhone 6` instead by the library mentioned above:

https://github.com/piwik/device-detector/blob/master/regexes/device/mobiles.yml#L200

Mask:
This is what you can get out of piwik if you set the user-agent appropriately: 

screen shot 2015-01-23 at 14 04 25

tarbrain:
A similar feature has just been merged into the android version of the piwik sdk:

https://github.com/piwik/piwik-sdk-android/pull/15

tarbrain:
Hello @mattiaslevin 

Have you yet come to a decision regarding this pull request? We have our fork in production for several months now and all looks very good.

It would be very nice to see a new release of the SDK containing all recent changes soon so we can switch from our fork back to the official repo.

Best Regards
Julian Krumow

mattab:
Hi @tarbrain !

@mattiaslevin and Piwik team are looking for a new maintainer to the iOS project as Mattias has got a new job and won't have ability to do much work on the project. Maybe you'd be interested to participate?

tarbrain:
Hello @mattab,

thank you for the offer but I am afraid I have to refuse.

We have forked the SDK to get the one or other missing feature into the code base - which worked pretty well by the way - but I do not have the time to fill the role as a full-time maintainer.

Nevertheless we would be very happy if you could merge our changes ;-)

Best regards and keep it up!
Julian Krumow

mattiaslevin:
Changes merged with some small change
* setUserAgent is optional to not break any users existing dispatchers
* The download feature have already been implemented.

Thanks a lot.

tarbrain:
Thank you. This is great!

PR #76, opened 2015-10-06
load MOM from PiwikTracker class's bundle

bgerstle:
Prevents exceptions when integrated as a framework (e.g. using CocoaPods with `use_frameworks!`).


mattiaslevin:
Great, nice catch.

Issue #50, opened 2014-10-20
Migrate to Swift


mattiaslevin:
Rewrite the SDK in Swift.

Erulezz:
+1 

Would really like to see this.

pjc-is:
+1

that would be awesome!

mattiaslevin:
Ok, I will start on a Swift implementation.
If anyone care to help out please let me know.

Erulezz:
If you need help with features or testing let me know!

funky-monkey:
+10

mattiaslevin:
ok ok, I have started and have something, lets see if I can get it to a shareable state.

Urkman:
What is the status here? How can I use Piwik in my Swift Project?

Erulezz:
You can use this Objective-C project in a Swift project with a bridging header. More info;

https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html

perlguy99:
I'd love to see this too.  How far along are you?  Need any help?  I may be able to help with it, if needed.

mattab:
Hi @perlguy99 @Erulezz @Urkman @funky-monkey - your help would be very welcome! @mattiaslevin is very busy at the moment and we would welcome new maintainers and help for the official Piwik iOS SDK :+1: if you are interested please consider opening pull requests. We could also give you commit access once you have contributed some things  :-) Thank you for your consideration!

mattab:
FYI:  we the Piwik team are looking for a maintainer for the iOS SDK. if you can help please comment here #91 - we need you!

JekaNaloyko:
I'll need to refactor the SDK to Swift, as I don't want to use AFNetworking. Already using Alamofire. Will comment here after prototyping the approach.

SteamGorilla:
Hi @mattab what's up about the Swift portage for piwik ?

mattab:
@SteamGorilla we are still looking for a volunteer to step up and take over the Piwik iOS SDK project. Would anyone be interested? 

It's a unique opportunity to help participate in a successful open source project and make a difference. We posted on twitter recently asking for help: 
![ios](https://cloud.githubusercontent.com/assets/466765/18828591/0cde4754-8435-11e6-808e-eb2f11e46acc.png)

SteamGorilla:
@mattab Okay, thanks for your answer. I'm not already skilled enough to realize a portage by myself :/

[...]

PR #96, opened 2016-10-15
[WIP] Swift 3 Port

brototyp:
I started to port the sdk over to Swift 3 (see #50). This is just the groundwork, but I would love to start a conversation early to try to get the API as familiar as possible. 

Core things I wanted to tackle:
- Prevent having a single big file or class responsible for everything.
- The objc SDK featured a protocol for different dispatchers. I wanted to do the same to queue events.
- Improve testability throughout the codebase.

Core questions I would like to discuss:
- SessionStart was a property before. Wouldn't it be better to be a method? It feels like it is an action to start a new session.
- The documentation states that custom variables are deprecated (http://piwik.org/docs/custom-variables/). Is it best to only implement custom dimensions?
- There are some features that are only supported via a token_auth (http://developer.piwik.org/api-reference/tracking-api). Those features weren't supported in the objc sdk. Is there a reason for that? What would be the best way to support those?
- The url property is necessary for every event. What is the best value for this property? The android sdk uses http://(appname) as far as I can see. The objc sdk uses http://(appname)/(action) and sometimes simply the last tracked url. Is there a necessity to track anything but http://(appname) ?
- How far is it fine to change the API of the SDK? It is possible to keep the API (mostly) the same as with the objc API. But I would like to change some pieces. How open are you to changes? In the end it is a major update (I guess) and people expect to change some bits and pieces. On the other hand, the old SDK will not be supported anymore (right?) so we might make the migration as easy as possible.

Looking forward to your feedback.


mattiaslevin:
Maybe if you create a Swift3 branch and do your PR towards that?
This way you can do smaller PRs that can be merged and a little easier for others to contribute :)

mattiaslevin:
I will leave a few comments, hope that is ok :)

PiwikTracker-Swift/Campaign.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  Campaign.swift
+//  PiwikTracker
+//
+//  Created by Cornelius Horstmann on 09.10.16.
+//  Copyright © 2016 Mattias Levin. All rights reserved.
```
mattiaslevin: I would probably remove these comments completely and just add descriptions for classes/structs when necessary.

PiwikTracker-Swift/Campaign.swift
```diff
@@ -0,0 +1,41 @@
...
+    let url: URL
+    
+    init?(_ string: String) {
+        if let url = URL(string: string) {
+            self.init(url)
+        }
+        return nil
+    }
+    
+    init?(_ url: URL) {
```
mattiaslevin: Maybe a descriptive parameter name

PiwikTracker-Swift/Campaign.swift
```diff
@@ -0,0 +1,41 @@
...
+        let urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        if let queryItems = urlComponents?.queryItems {
+            let nameItem = queryItems.filter({$0.name == PiwikConstants.URLCampaignName}).first
```
mattiaslevin: You can use `first { $0.name == PiwikConstants.URLCampaignName }`

PiwikTracker-Swift/CustomVariable.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+
+import Foundation
+
+@objc public enum CustomVariableScope: Int {
+    case Visit
+    case Screen
+}
+
+class CustomVariable {
```
mattiaslevin: Maybe `final`?

PiwikTracker-Swift/CustomVariable.swift
```diff
@@ -0,0 +1,28 @@
...
+//  CustomVariable.swift
+//  PiwikTracker
+//
+//  Created by Cornelius Horstmann on 09.10.16.
+//  Copyright © 2016 Mattias Levin. All rights reserved.
+//
+
+import Foundation
+
+@objc public enum CustomVariableScope: Int {
```
mattiaslevin: Could consider moving the enum inside the class below ans just rename it to `Scope`
(you will get some namespace help)

PiwikTracker-Swift/Dictionary+Extension.swift
```diff
@@ -0,0 +1,33 @@
...
+                filtered[k] = value
+            }
+        }
+        return filtered
+    }
+    
+}
+
+extension Array {
+    public func strings() -> [String] {
```
mattiaslevin: Considering how you are using this method further below I suggest you remove this and do
`components.flatMap { $0 }.joined(separator: "/")`
to remove `nil` values.

PiwikTracker-Swift/CustomVariable.swift
brototyp: That's a great idea! Haven't thought of that.

PiwikTracker-Swift/Dictionary+Extension.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+
+import Foundation
+
+protocol OptionalType {}
+
+extension Optional: OptionalType {}
+
+extension Dictionary {
+    static func removeOptionals(dictionary: [Key: Value?]) -> [Key:Value] {
```
mattiaslevin: Since this is a static method I would probably not put in an extension.
The alternative would be to make this a generic instance method.

brototyp:
> Maybe if you create a Swift3 branch and do your PR towards that?

Should we better close this PR here then and chunk it off into smaller ones?

mattab:
Cheers @mattiaslevin for providing @brototyp early feedback. that's very appreciated :sunglasses:

PiwikTracker-Swift/UIDevice+Extension.swift
```diff
@@ -0,0 +1,98 @@
...
+        // FIXME: implement me
+        return CGSize(width: 0, height: 0)
+        }}
+    
+    static var osVersion: String { get {
+        // FIXME: distinct between iOS and Mac OS
+        return UIDevice.current.systemVersion
+        }}
+    
+    static var appVersion: String { get {
```
thorstenstark: Wouldn't that fit better in UIApplication+Extension.swift?

brototyp:
We will keep this branch for now, for further reference.

Issue #23, opened 2013-09-15
Restructure workspace and project


mattiaslevin:
Restructure workspace and project

```
Piwik (workspace)
  PiwikTracker (folder)
    source code
  PiwikiOSDemo (project with an iOS build target)
    demo source code
  PiwikOXDemo (project with and OSX build target
    demo source code
  Pods (pod dependency targets)
```

Look at the AFNetworking organisation.

Need different demo project and/or test project for iOS and OSX

brototyp:
Implemented.

PR #97, opened 2016-10-24
restructured workspace and removed objc framework code

brototyp:
This commit sets the base for rewriting the sdk in swift 3. Since I was restructuring most of the workspace I figured I can consider #23 aswell.
This is why I moved a bunch of files from a to b. I hope this matches your vision @mattiaslevin.

The structure now is:

```
- PiwikTracker.xcworkspace
- PiwikTracker.xcodeproject
- PiwikTracker/
  - // this is where the framework code will live
- Example/
  - // this is where all the example apps will live
  - Example.xcodeproj
  - iOS Example/
  - // we could add the `macOS/` folder here
```


.gitignore
```diff
@@ -15,10 +15,9 @@ build/*
 *.perspective
 *.perspectivev3
 *.pbxuser
-*.xcworkspace
```
brototyp: The workspace now integrates the framework and the example app. So we should include it.

mattiaslevin:
I created and pushed a `swift3` branch.
Can you please change the target of you PR this branch instead of master?

brototyp:
Yes of cause. Done.

minitrue22:
I think we should go full out with Swift3.
Suggest creating an empty project using the package manager
https://swift.org/getting-started/#using-the-package-manager
and copy over code later on if needed

brototyp:
Hi @minitrue22,

I am not sure about that. I would love to offer at least a CocoaPods Podspec as well. How would that be the case? What is, beside using the newest technology, the upside in using SPM in this case?

minitrue22:
I do not think there is anything excluding support for Cocoapods and Carthage. Its more about which tools default project setup should be used :)

It should be possible to create a Cocoapod from the folder structure created by package manager later on.

My motivation goes something like this:
-This library is written in Swift3
-Swift3 is not binary compatible with Swift2
-Cocoapods and Carthage was more or less the only option for dependency mgmt before Swift3
-When developers rewrite their frameworks and apps to Swift3 (because it is a rewrite, changes are big) many/(most?) will add support for package manager. Supporting Cocoapods will not give access to the bulk of Swift pods since they are primarily in Swift2.

I also think that Cocoapods only chance for survival long term in the Swift3 echo system is to offer some kind of integration with package manager.

brototyp:
Ok, I can see the point in using SPM and defining it (for now) as a default. I will check the documentation and will have a look on how SPM and "regular projects", created with Xcode work together.

brototyp:
I now researched a bit about the current state of the SPM. I see one major downside. The SPM is by fare not as mature as other package managers. So there are things that are simply not possible (for now). One example is dependencies that are there just for the test target (I love to use Quick and Nimble). Additionally there are some things that are still open and discussed. There is no `best way` for many things such as: Where to put the example application.

The question I would like to ask: What is the downside to start with the swift-lib as is, ignoring the SPM, and revisiting this discussion regularly? In the end all the code is the same.

brototyp:
I decided to merge this branch for now to let the Swift rewrite progress again. In the end there should not be any issue to adding SPM files.

PR #101, opened 2016-12-18
added PiwikUserDefaults

brototyp:
I figured a global Constants file is neither necessary nor good. I moved all UserDefaults related constants in a Constants struct in the PiwikUserDefaults.
I am not sure about the naming an whether we should omit the `Piwik` prefix in local classes. In this case it would be a bit complicated because of the conflict to the `UserDefauts`.

### DIFF
--- a/PiwikTracker/PiwikUserDefaults.swift
+++ b/PiwikTracker/PiwikUserDefaults.swift
@@ -4,91 +4,77 @@ import Foundation
 /// mapping onto values stored in the UserDefaults.
 /// All getter and setter are sideeffect free and automatically syncronize
 /// after writing.
-internal class PiwikUserDefaults {
-    internal static let standard = PiwikUserDefaults()
-    internal let userDefaults = UserDefaults.standard
+internal struct PiwikUserDefaults {
+    static let standard = PiwikUserDefaults()
+    let userDefaults = UserDefaults.standard
     
-    internal var totalNumberOfVisits: UInt32 {
+    var totalNumberOfVisits: Int {
         get {
-            return UInt32(self.userDefaults.integer(forKey: PiwikUserDefaults.Constants.UserDefaultTotalNumberOfVisitsKey))
+            return userDefaults.integer(forKey: PiwikUserDefaults.Key.totalNumberOfVisits)
         }
         set {
-            self.userDefaults.set(newValue, forKey: PiwikUserDefaults.Constants.UserDefaultTotalNumberOfVisitsKey)
-            self.userDefaults.synchronize()
+            userDefaults.set(newValue, forKey: PiwikUserDefaults.Key.totalNumberOfVisits)
+            userDefaults.synchronize()
         }
     }
     
-    internal var firstVisit: Date? {
+    var firstVisit: Date? {
         get {
-            return self.date(forKey: PiwikUserDefaults.Constants.UserDefaultFirstVistsTimestampKey)
+            return userDefaults.object(forKey: PiwikUserDefaults.Key.firstVistsTimestamp) as? Date
         }
         set {
-            self.userDefaults.set(newValue, forKey: PiwikUserDefaults.Constants.UserDefaultFirstVistsTimestampKey)
-            self.userDefaults.synchronize()
+            userDefaults.set(newValue, forKey: PiwikUserDefaults.Key.firstVistsTimestamp)
+            userDefaults.synchronize()
         }
     }
     
-    internal var previousVisit: Date? {
+    var previousVisit: Date? {
         get {
-            return self.date(forKey: PiwikUserDefaults.Constants.UserDefaultPreviousVistsTimestampKey)
+            return userDefaults.object(forKey: PiwikUserDefaults.Key.previousVistsTimestamp) as? Date
         }
         set {
-            self.userDefaults.set(newValue, forKey: PiwikUserDefaults.Constants.UserDefaultPreviousVistsTimestampKey)
-            self.userDefaults.synchronize()
+            userDefaults.set(newValue, forKey: PiwikUserDefaults.Key.previousVistsTimestamp)
+            userDefaults.synchronize()
         }
     }
     
-    internal var currentVisit: Date? {
+    var currentVisit: Date? {
         get {
-            return self.date(forKey: PiwikUserDefaults.Constants.UserDefaultCurrentVisitTimestampKey)
+            return userDefaults.object(forKey: PiwikUserDefaults.Key.currentVisitTimestamp) as? Date
         }
         set {
-            self.userDefaults.set(newValue, forKey: PiwikUserDefaults.Constants.UserDefaultCurrentVisitTimestampKey)
-            self.userDefaults.synchronize()
+            userDefaults.set(newValue, forKey: PiwikUserDefaults.Key.currentVisitTimestamp)
+            userDefaults.synchronize()
         }
     }
     
-    internal var optOut: Bool {
+    var optOut: Bool {
         get {
-            return self.userDefaults.bool(forKey: PiwikUserDefaults.Constants.UserDefaultOptOutKey)
+            return userDefaults.bool(forKey: PiwikUserDefaults.Key.optOut)
         }
         set {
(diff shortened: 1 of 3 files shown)


PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
+import Foundation
+
+/// PiwikUserDefaults is a wrapper for the UserDefaults with properties
+/// mapping onto values stored in the UserDefaults.
+/// All getter and setter are sideeffect free and automatically syncronize
+/// after writing.
+internal class PiwikUserDefaults {
```
mattiaslevin: I recommend to always mark classes as `final` unless they should be subclasses (performance reason)

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+    internal static let standard = PiwikUserDefaults()
```
mattiaslevin: I would not use any access modifiers for members unless they should have a more restrictive modifier then the enclosing type. Less clutter and typing.
Style question, up to you.

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+    internal let userDefaults = UserDefaults.standard
+    
+    internal var totalNumberOfVisits: UInt32 {
+        get {
+            return UInt32(self.userDefaults.integer(forKey: PiwikUserDefaults.Constants.UserDefaultTotalNumberOfVisitsKey))
```
mattiaslevin: I suggest only to use `self.` where enforced by the compiler. Less clutter and typing.
Style question, up to you.

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+
+/// PiwikUserDefaults is a wrapper for the UserDefaults with properties
+/// mapping onto values stored in the UserDefaults.
+/// All getter and setter are sideeffect free and automatically syncronize
+/// after writing.
+internal class PiwikUserDefaults {
+    internal static let standard = PiwikUserDefaults()
+    internal let userDefaults = UserDefaults.standard
+    
+    internal var totalNumberOfVisits: UInt32 {
```
mattiaslevin: I would probably return `Int`.
Swift guide suggest using `Int` in most cases.

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+        get {
+            return self.userDefaults.bool(forKey: PiwikUserDefaults.Constants.UserDefaultOptOutKey)
+        }
+        set {
+            self.userDefaults.set(newValue, forKey: PiwikUserDefaults.Constants.UserDefaultOptOutKey)
+            self.userDefaults.synchronize()
+        }
+    }
+    
+    internal func clientId(prefix: String) -> String? {
```
mattiaslevin: What is the `prefix` used for?

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+    
+    internal func set(clientId: String, prefix: String) {
+        let key = "\(prefix)_\(PiwikUserDefaults.Constants.UserDefaultVisitorIDKey)"
+        self.userDefaults.setValue(clientId, forKey: key)
+        self.userDefaults.synchronize()
+    }
+}
+
+extension PiwikUserDefaults {
+    internal func set(date: Date, forKey key: String) {
```
mattiaslevin: This method is never used and can probably be removed.
`userDefaults.set(...)` accepts `NSData` as value.  Not sure there is a need to cover it into a timestamp before saving.

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+        let timestamp = date.timeIntervalSince1970
+        self.userDefaults.set(timestamp, forKey: key)
+    }
+    internal func date(forKey key: String) -> Date? {
+        guard self.userDefaults.object(forKey: key) != nil else { return nil }
```
mattiaslevin: If you save the `NSDate` in default it should be enough if you do
````
return self.userDefaults.object(forKey: key) as? Date
```

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+    
+    internal func set(clientId: String, prefix: String) {
+        let key = "\(prefix)_\(PiwikUserDefaults.Constants.UserDefaultVisitorIDKey)"
+        self.userDefaults.setValue(clientId, forKey: key)
+        self.userDefaults.synchronize()
+    }
+}
+
+extension PiwikUserDefaults {
```
mattiaslevin: I guess you could make this and extension to `UserDefaults`

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+    internal func date(forKey key: String) -> Date? {
+        guard self.userDefaults.object(forKey: key) != nil else { return nil }
+        let storedDouble = self.userDefaults.double(forKey: key)
+        return Date(timeIntervalSince1970: storedDouble)
+    }
+}
+
+extension PiwikUserDefaults {
+    fileprivate struct Constants {
+        static let UserDefaultTotalNumberOfVisitsKey = "PiwikTotalNumberOfVistsKey"
```
mattiaslevin: all parameters should start with small char (also constants)

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+    internal func date(forKey key: String) -> Date? {
+        guard self.userDefaults.object(forKey: key) != nil else { return nil }
+        let storedDouble = self.userDefaults.double(forKey: key)
+        return Date(timeIntervalSince1970: storedDouble)
+    }
+}
+
+extension PiwikUserDefaults {
+    fileprivate struct Constants {
```
mattiaslevin: I would probably move the struct inside the class, maybe rename it to `Key`.
I would also consider removing the `UserDefault` from the variable name.

This will make it short and nice `PiwikUserDefaults.Key.totalNumberOfVisitsKey`

PiwikTracker/PiwikUserDefaults.swift
mattiaslevin: Or even better use a struct.

PiwikTracker/PiwikUserDefaults.swift
brototyp: Good idea. I changed it to `PiwikUserDefaults.Key.totalNumberOfVisits`

brototyp: > I would probably move the struct inside the class

I usually like to put those things "away" into an extension since nobody really has to bother about the values.

PiwikTracker/PiwikUserDefaults.swift
brototyp: True. Removed both method and stored the date directly.

PiwikTracker/PiwikUserDefaults.swift
brototyp: Changed it to a struct.

PiwikTracker/PiwikUserDefaults.swift
brototyp: True. Removed.

PiwikTracker/PiwikUserDefaults.swift
brototyp: The `prefix` is an old piece of code meant to enable two PiwikTracker for two different backends with two different site ids. So as a prefix the site id could be used.

I removed it for now.

brototyp:
One thing I did not think about up until now:

What if one wants to use two different PiwikTracker with two different backends. This way the values in the PiwikUserDefaults could be different. What do you think about instantiating the PiwikUserDefaults with a prefix (which then could be the site id) and prefixing every userdefaults key with it.

brototyp:
I postponed the discussion about multiple PiwikTracker instances to here: #118

Issue #102, opened 2016-12-19
UIDevice Extension


brototyp:
An extension to return:

- device platform (iPhone1,1)
- platform name (iPhone 1G)
- screen size 
- version of the operating system.

Discussion: Is an extension the best idea? There might be conflicts with the main application in which the piwik-sdk is used. I guess it is better to use a separate class.

thorstenstark:
Yeah, moving all that stuff, including the extensions from UIApplication, to a seperate class sounds reasonable. 
What do you think about "AppContext" as name for this class?

brototyp:
I am not sure if I would move all those methods into one class. If you want to do that i would name it just `Context` because the device platform doesn't really have anything to do with the Application itself. An alternative would be:

```
class Context {
  class Application {}
  class Device {}
}
```

... with all methods inside are class methods. This way everything is the same (the Context), is different (Device, Application) and is still easily accessible (Context.Application.version or Context.Device.platform).

thorstenstark:
That's a good idea.

PR #104, opened 2016-12-19
added Extension for UIDevice #102

thorstenstark:
I'm not sure if the screen sizes fit in this extension, as both use the UIScreen class to get the informations.


brototyp:
> I'm not sure if the screen sizes fit in this extension, as both use the UIScreen class to get the informations.

Yeah. I understand that. Have you seen my thought over in the #102 ? Maybe the best idea is to move all those method in a separate Class `Device`. Therefore we don't fiddle around with an existing class and we can keep the screen size method/property.

thorstenstark:
As mentioned in the #102 I moved these methods to a Context class.

PiwikTracker/Context.swift
```diff
@@ -0,0 +1,115 @@
...
+        }()
+        
+        // Returns the screen size in points
+        static let screenSize: String = {
+            let bounds = UIScreen.main.bounds
+            return "\(bounds.size.width)x\(bounds.size.height)"
+        }()
+        
+        // Returns the screen size in pixels
+        static let physicalScreenSize: String = {
```
brototyp: I would rather like to name it nativeScreenSize to match it with the `nativeBounds`. For physicalScreenSize I would expect it to be in centimeters or inch.

mattab:
> Can you squash all commits into one?

@brototyp Fyi Github now has this feature to let you squash all commits with the UI. So usually it's not needed to ask contributors to squash. Learn more here: https://github.com/blog/2141-squash-your-commits

brototyp:
> Fyi Github now has this feature to let you squash all commits with the UI.

Yeah, right. Forgot that. I rarely use GitHub.

PiwikTracker/Context.swift
thorstenstark: Good point. Changed it.

PiwikTracker/Context.swift
```diff
@@ -0,0 +1,115 @@
+import UIKit
+
+public class Context {
```
mattiaslevin: imho, having nested classes with only static members (all internally available) does not add any value.
We are using namespace to avoid clashes and `Device` and `Application` does not share any `fileprivate` code not visible to the outside.

I would just make a stand-alone `Device` struct (prefer struct over class, or make class `final`)

PiwikTracker/Context.swift
```diff
@@ -0,0 +1,115 @@
...
+    
+    public class Application {
+        // TODO: implement me
+    }
+    
+    public class Device {
+        
+        /// The platform name of the device i.e. "iPhone1,1" or "iPad3,6"
+        static let platform: String = {
```
mattiaslevin: Using static methods make access to members more easy without passing around an instance.
However it is a bit like having singletons and might bite back if you ever need to mock the class in a unit test.

But I do not think that will be an issue here, so I would keep as is and refactor later the need arise.

PiwikTracker/Context.swift
```diff
@@ -0,0 +1,115 @@
+import UIKit
+
+public class Context {
+    
+    public class Application {
+        // TODO: implement me
+    }
+    
+    public class Device {
```
mattiaslevin: I would make this a struct

mattiaslevin: Consider asking access control (internal) to the top level type to be consistent with the other PR.
I think this code style guide is pretty good.
https://github.com/raywenderlich/swift-style-guide
https://github.com/github/swift-style-guide

PiwikTracker/Context.swift
thorstenstark: So we make two structs 'Device' and 'Application' not nested in a 'Context'?! What do you think @brototyp ?

brototyp: The only value nesting both into a "Context" is for the developer. This way we declare both as being some Kind of Context and grouping both together. I am fine with both.

PiwikTracker/Context.swift
brototyp: Good point. So refactoring wise we should add something like a `current` singleton which then returns an Object with all the current values. This way we cloud instantiate a Device with mocked values.

PiwikTracker/Context.swift
thorstenstark: In this case I would go with the stand-alone `Device` struct.

PiwikTracker/Context.swift
thorstenstark: I will make a proposal for the variant with the singleton.

PiwikTracker/Context.swift
```diff
@@ -0,0 +1,115 @@
...
+            case "iPad4,9":      return "iPad Mini 3 (A1601)"
+            case "iPad5,1":      return "iPad Mini 4 (WiFi)"
+            case "iPad5,2":      return "iPad Mini 4 (Cellular)"
+            case "iPad5,3":      return "iPad Air 2 (WiFi)"
+            case "iPad5,4":      return "iPad Air 2 (Cellular)"
+            case "iPad6,3":      return "iPad Pro 9.7 (WiFi)"
+            case "iPad6,4":      return "iPad Pro 9.7 (Cellular)"
+            case "iPad6,7":      return "iPad Pro 12.9 (WiFi)"
+            case "iPad6,8":      return "iPad Pro 12.9 (Cellular)"
+
```
thorstenstark: Is the Piwik SDK supposed to run on WatchOS and tvOS, too? Then we should add the proper device names here.

thorstenstark:
I added a version with a `current` singleton that holds the properties for the current device.

PiwikTracker/Device.swift
```diff
@@ -0,0 +1,132 @@
...
+                                       humanReadablePlatformName: currentHumanReadablePlatformName(),
+                                       osVersion: currentOSVersion(),
+                                       screenSize: currentScreenSize(),
+                                       nativeScreenSize: currentNativeScreenSize())
+    /// The platform name of the device i.e. "iPhone1,1" or "iPad3,6"
+    let platform: String
+    
+    /// A human readable version of the platform name i.e. "iPhone 6 Plus" or "iPad Air 2 (WiFi)"
+    /// Or the platform name in case no human readable name was found.
+    let humanReadablePlatformName: String
```
brototyp: The human readable name could be a function that calculates the human readable version from the platform. And it could return an optional that is nil if no human readable version exists.

PiwikTracker/Device.swift
```diff
@@ -0,0 +1,132 @@
...
+    
+    /// The version number of the OS as String i.e. "1.2" or "9.4"
+    let osVersion: String
+    
+    // The screen size in points
+    let screenSize: String
```
brototyp: I think the size should be a CGSize.

PiwikTracker/Device.swift
```diff
@@ -0,0 +1,132 @@
...
+    private static func currentScreenSize() ->  String {
+        let bounds = UIScreen.main.bounds
+        return "\(bounds.size.width)x\(bounds.size.height)"
+    }
+    
+    // Returns the screen size in pixels
+    private static func currentNativeScreenSize() ->  String {
+        let bounds = UIScreen.main.nativeBounds
+        return "\(bounds.size.width)x\(bounds.size.height)"
+    }
```
brototyp: Those static methods with a `current` prefixed to irritate me a bit. A possible alternative would be a static method on `Device` that returns a new Device populated with values for the current device. What do you think?

PiwikTracker/Device.swift
```diff
@@ -0,0 +1,132 @@
...
+    let humanReadablePlatformName: String
+    
+    /// The version number of the OS as String i.e. "1.2" or "9.4"
+    let osVersion: String
+    
+    // The screen size in points
+    let screenSize: String
+    
+    // The screen size in pixels
+    let nativeScreenSize: String
```
brototyp: This should be a CGSize as well.

PiwikTracker/Context.swift
brototyp: That's a good idea. Yes add them please.

PiwikTracker/Device.swift
```diff
@@ -0,0 +1,132 @@
+import UIKit
+
+internal struct Device {
+    
+    /// A shared instance holding the properties for the current device
+    static let current:Device = Device(platform: currentPlatform(),
```
minitrue22: I would suggest removing the static `current` and have a factory method that returns the current device
````
static func makeCurrentDevice() ->  Device {
  let platform = currentPlatform()
  ....
   return Device(platform: platform, ...)
}
````

I would only create a global shared instance later on if it becomes needed.
Most probably the "tracker class" can have a simple instance variable it can instantiate at construction by calling the static method above (reducing coupling and making testing more simple).

File: PiwikTracker/Device.swift (hunk 6 above)
Comment (minitrue22) [inline PiwikTracker/Device.swift]: Agree that returning an optional String is probably a good thing.
If the caller really needs a string it can do 
````
let platformName = device. humanReadablePlatformName ?? device = platform 
````
But that is the callers responsibility not this struct.

File: PiwikTracker/Device.swift (hunk 8 above)
Comment (minitrue22) [inline PiwikTracker/Device.swift]: I think a factory method would be great.
Even so, I think it is good practice to write small functions. The methods above does exactly what they promise :)

File: PiwikTracker/Device.swift (hunk 7 above)
Comment (thorstenstark) [inline PiwikTracker/Device.swift]: Agree.

File: PiwikTracker/Device.swift (hunk 9 above)
Comment (thorstenstark) [inline PiwikTracker/Device.swift]: Agree.

File: PiwikTracker/Device.swift (hunk 10 above)
Comment (thorstenstark) [inline PiwikTracker/Device.swift]: Done.

File: PiwikTracker/Device.swift (hunk 8 above)
Comment (thorstenstark) [inline PiwikTracker/Device.swift]: I changed the method names a bit to remove irritation ;)

File: PiwikTracker/Device.swift [hunk 11]
```diff
@@ -0,0 +1,154 @@
+import UIKit
+
+internal struct Device {
+    
+    /// Creates an returns a new device object representing the current device
+    static func makeCurrentDevice() ->  Device {
```
brototyp: Not overly happy with the word make here but it's not a huge issue.

PiwikTracker/Device.swift
brototyp: I like that!

PR #113, opened 2016-12-21
added Event model

brototyp:
This is the first idea for an Event Model. One important thing to discuss is how to add the other properties for Actions, Ecommerce and such.

1. More Parameters: We could simply add more and more parameters. This way the parameters could really add up over time.
2. Subclassing: We could change the Event to be a class and then use Subclassing to define specialized Events such as an Ecommerce Event.
3. Sub-Structs: We could create a Struct for all Ecommerce properties and then reference those in the Event. 
4. Any other Ideas?

What do you think @thorstenstark ?


PiwikTracker/Event.swift
```diff
@@ -0,0 +1,81 @@
...
+    internal let firstVisit: Date?
+    
+    /// The language of the device.
+    /// Should be in the format of the Accept-Language HTTP header field.
+    /// api-key: lang
+    internal let language: String
+    
+    /// Should be set to true for the first event of a session.
+    /// api-key: new_visit
+    internal let isNewSesssion: Bool
```
thorstenstark: One s too much.

PiwikTracker/Event.swift
```diff
@@ -0,0 +1,81 @@
...
+/// - All Plugins: fla, java, dir, qt, realp, pdf, wma, gears, ag
+/// - cid: We will use the uid instead of the cid.
+/// - dimension: For now we don't support dimension.
+/// - UserAgent (ua) should be set in the dispatcher, right?
+internal struct Event {
+    internal let uuid = NSUUID().uuidString
+    
+    /// The Date and Time the event occurred.
+    /// api-key: h, m, s
+    internal let date: Date
```
thorstenstark: Will there be cases where date is another date than now? If not, what about let date = Date() ?

PiwikTracker/Event.swift
```diff
@@ -0,0 +1,81 @@
...
+//
+//  Created by Cornelius Horstmann on 21.12.16.
+//  Copyright © 2016 PIWIK. All rights reserved.
+//
+
+import Foundation
+
+/// Represents an event of any kind.
+///
+/// - Note: Should we store the resolution in the Event (cleaner) or add it before transmission (smaller)?
```
thorstenstark: I would put add the resolution the same place where we add the user agent.

PiwikTracker/Event.swift
mattiaslevin: I would consider removing access control info on members.

mattiaslevin:
I think this is hard to know upfront. 
Maybe better to implement a basic event only containing the properties needed for the most basic use case (I guess that is screen/page view). Make a good solution that you like that solves the problem.

When adding the next use case, e.g. action, refactor to make it fit. Either inheritance or composition (action event can have a base event as a property).

PiwikTracker/Event.swift
```diff
@@ -0,0 +1,81 @@
...
+    internal let actionName: [String]
+    
+    /// Unique ID per visitor (device in this case). Should be
+    /// generated upon first start and never changed after.
+    /// api-key: _id
+    internal let visitorId: String
+    
+    /// An optional user identifier such as email or username.
+    /// api-key: uid
+    internal let userId: String?
```
mattiaslevin: I would consider putting some of the properties into their own struct, e.g. "user related", "session related", "device related" etc.

If you are lucky you can avoid recreating some of these structs for each event, e.g. device and user related properties does not change and can be instantiated once only. Session properties only need to be reset when the session is reset etc.

PiwikTracker/Event.swift
brototyp: Yep. For example if you serialize and de-serialize an Event.

PiwikTracker/Event.swift
brototyp: Hm. Unsure about that. In the end the useragent represents the piece of code responsible for transmitting the data. Per default I would like it to be defined by the transmitting object/class/struct.

The resolution is independent of the transmitter and somehow it feels wrong to add it there. 

Idea: What if the Event has a reference to a "Context" object that is a reference to a Singleton (Context.current or such). This way we don't have to serialize/de-serialize it for storing it in CoreData and such but it is still a property of the Event.

PiwikTracker/Event.swift
brototyp: Yep

brototyp:
> Maybe better to implement a basic event only containing the properties needed for the most basic us ...

I thought about the same. This way we can split of the PiwikTracker-Issue into smaller issues implementing certain types of tracking (e.g. Download or such) and the decide how to add/store additional properties.

PiwikTracker/Event.swift
brototyp: :+1: fixed

PiwikTracker/Event.swift
```diff
@@ -0,0 +1,87 @@
...
+    let userId: String?
+}
+
+internal struct Session {
+    /// The number of sessions of the current user.
+    /// api-key: _idvc
+    let sessionsCount: Int
+    
+    /// The timestamp of the previous visit.
+    /// Discussion: Should this be now for the first request?
```
thorstenstark: I would expect it to bel `nil` for the first visit, because there never was a previous visit. Same for the firstVisit. But having it a non optional would be preferrable.
In the documentation I couldn't find anything about what Piwik expects it to be in this case (now or nil or whatever). Perhaps now would also be suitable and has the same effect for the statistics.

brototyp: @mattab What should the values for _viewts and _idts be for the first session/visit/request? Is there any reason to not set it to the current date and time?

minitrue22: I would make a simplification and make them non optional for now.
The first time set them to same value as "current visit".
If this turns out to be an issue refactor later.

PiwikTracker/Event.swift
```diff
@@ -0,0 +1,87 @@
...
+    let actionName: [String]
+    
+    /// The language of the device.
+    /// Should be in the format of the Accept-Language HTTP header field.
+    /// api-key: lang
+    let language: String
+    
+    /// Should be set to true for the first event of a session.
+    /// api-key: new_visit
+    let isNewSession: Bool
```
minitrue22: I am not sure if this should go into the `Session` struct.
Maybe just keep as is and refactor later.

minitrue22:
I think this can be merged soon.

PiwikTracker/Event.swift
brototyp: Refactored this to be a non-optional. Will keep it like this for now. Might refactor later.

mattab: > What should the values for _viewts and _idts be for the first session/visit/request? Is there any reason to not set it to the current date and time?

@brototyp It is fine to set _viewts and _idts on the first visit as well. Piwik Tracking API will use these timestamps to process a number of "days since", which will be zero, so it'll work as expected.

PR #115, opened 2016-12-30
Added the Queue and Queueable Protocol

brototyp:


mattiaslevin:
Difficult to provide good comments before you see how it will be used :)
I guess this will be used to store and retrieve events for sending?
Implementation can either be memory based or persisted somehow (core data or file)

I would consider providing one implementation in the PR.

You have a couple of standard queue implementation here.
https://github.com/raywenderlich/swift-algorithm-club

PiwikTracker/Queue.swift
```diff
@@ -0,0 +1,47 @@
...
+//  PiwikTracker
+//
+//  Created by Cornelius Horstmann on 09.10.16.
+//  Copyright © 2016 Mattias Levin. All rights reserved.
+//
+
+import Foundation
+
+/// An Element that is queueable.
+protocol Queueable {
```
mattiaslevin: I guess you want this so the queue implementation can pull out the data and store it?

PiwikTracker/Queue.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// A unique identifier. It should be unique per object.
+    var id: UUID { get }
+    
+    /// This property is used to get the data that should be stored in the queue.
+    var data: Data { get }
+    
+    /// This initializer is used when dequeueing.
+    init?(data: Data)
```
mattiaslevin: Not sure this will work. 
The queue implementation will not/should not know the original type is it sent, only that it can call `data`. It can only return objects of `Queueable` and the protocol should not care of how it is instantiated.

PiwikTracker/Queue.swift
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/// A queue in which Objects that conform to queueable can be stored.
+protocol Queue {
+    /// The Item type have to conform to the Queueable protocol.
+    associatedtype Item: Queueable
```
mattiaslevin: I would maybe try/start without the `Queueable` protocol and only add it later if needed.

Maybe Item can conform to `NSCoding`?
You should be able to do exactly the same as with the `Queueable` (but I think it will force the object to be a class)

PiwikTracker/Queue.swift
mattiaslevin: Not sure `Queueable` is a good description it has nothing to do with queuing. More about archiving or storing.

brototyp:
Hey @mattiaslevin, thanks for your time reviewing this!

> Difficult to provide good comments before you see how it will be used 😊
> I guess this will be used to store and retrieve events for sending?

I see your point. On the other hand it you will need more than one implementation of Queue for that I guess. I would think you would need the usage of the Queue. Which then would be to much I guess. I will add one in memory queue to this queue (since it is pretty small). 

Usagewise: I basically copied the usage over from the objc implementation. Every produced event will be stored in this queue. So the PiwikTracker will create an Event (which conforms to queueable) and will store it in the Queue via the queue method. 
Depending on the configuration another piece of code will dequeue items, try to transmit them and requeue if transmission failed.

PiwikTracker/Queue.swift
brototyp: Right now the Queue can hold only one type of objects. So this would have worked. I thought that is sufficient to store all events, but now that I am writing this I see that it cripples the queue and limits its usage. Say we want to have different types of events (download event, view event, ...). We might decide to have different objects for those. I will refactor the queue to store all multiple types of objects.

PiwikTracker/Queue.swift
brototyp: Yep, NSCoding is only possible for objects. I will start with NSCoding and we can extend this if we need more than just classes.

PiwikTracker/Queue.swift
```diff
@@ -0,0 +1,24 @@
...
+    ///   - item: The Item to queue.
+    ///   - completion: A closure to be called once queueing is completed.
+    func queue(item: NSCoding, completion: ()->())
+    
+    /// Dequeues and returns a certain amount of items.
+    ///
+    /// - Parameters:
+    ///   - limit: The maximum amount of items to dequeue. May return less if less are queued.
+    ///   - completion: A closure to be called once the elements are dequeued. The closure will be called with the dequeued items.
+    func dequeue(withLimit limit: Int, completion: (_ items: [Any])->())
```
mattiaslevin: I would avoid using `Any` if possible and make the protocol generic instead.

````
protocol Queue {    
    associatedtype T: NSCoding    
    var itemCount: Int { get }
    func queue(item: T, completion: ()->())
    func dequeue(withLimit limit: Int, completion: (_ items: [T])->())
    func deleteAll()
}

extension Queue {
    func dequeue(completion: (_ items: [T])->()) {
        dequeue(withLimit: 1, completion: completion)
    }
}
````

File: PiwikTracker/MemoryQueue.swift [hunk 5]
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/// The MemoryQueue is a **thread-unsave** in memory Queue.
+class MemoryQueue: Queue {
```
mattiaslevin: I would make this a struct (or add `final`)

mattiaslevin: If you make it a struct you need to add `mutating` to the protocol members

PiwikTracker/MemoryQueue.swift
```diff
@@ -0,0 +1,28 @@
...
+    private var items = [Data]()
+    
+    var itemCount: Int { get {
+        return items.count
+        }
+    }
+    
+    func queue(item: NSCoding, completion: ()->()) {
+        let data = NSKeyedArchiver.archivedData(withRootObject: item)
```
mattiaslevin: There is no need to serialise the items in this case, you can just store them as is if you make the class generic.

And the queue will return the same type as you put in.

````
/// The MemoryQueue is a **thread-unsave** in memory Queue.
struct MemoryQueue<E: NSCoding>: Queue {    
    typealias T = E    
    private var items: [E] = []
    ...    
}
````

File: PiwikTracker/Queue.swift [hunk 7]
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+/// A queue in which Objects that conform to queueable can be stored.
```
mattiaslevin: Nitpicking, I think you can skip much of the comments and api documentation for internal classes. I would only add api documentation to types and methods visible outside the framework.

It is pretty clear what a queue does and the method names are self explanatory.

Comments have a tendency to root and become lies after awhile (already now this comment claim object must conform to queueable which is no longer true :)

PiwikTracker/MemoryQueue.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/// The MemoryQueue is a **thread-unsave** in memory Queue.
```
mattiaslevin: 👍 this is a good comment.

Not sure if "not thread safe" is a little clearer.

PiwikTracker/Queue.swift
mattiaslevin: You could even skip the type constraint completely and only do
````
associatedtype T
````
And let implementing types add any type constraints. E.g. in-memory implementations does not need any constraints but a persistent queue could require confirmation to `NSCoding`.

But the current limitation  does not matter much since the implementation will not be reused outside the framework.

brototyp: > I would avoid using Any if possible and make the protocol generic instead.

As far as I understand this way we aren't able to create a queue object that can store objects of multiple types. For example ViewEvent, DownloadEvent and others. In the end we do want to create a queue that can store Any objects.

brototyp: Oh, sure. I am wrong here. Of cause we then can instantiate an object of MemoryQueue<Any> to do that.

PiwikTracker/MemoryQueue.swift
brototyp: True *facepalm*.

PR #119, opened 2017-01-08
WIP: added dispatcher protocol

brototyp:
Issue: #106 

How should we handle Events that will never be transmitted. For example Events that are to old. It is not possible to submit events that are older than 24h. Should there be an option for the dispatcher to mark an event as **We will never be able to transmit this event, don't retry**?.

### DIFF
--- a/PiwikTracker/Dispatcher.swift
+++ b/PiwikTracker/Dispatcher.swift
@@ -1,0 +1,9 @@
+import Foundation
+
+internal protocol Dispatcher {    
+    var userAgent: String? { get }
+    
+    func send(event: Event, success: ()->(), failure: (_ shouldContinue: Bool)->());
+    
+    func send(events: [Event], success: ()->(), failure: (_ shouldContinue: Bool)->());
+}
--- a/PiwikTracker.xcodeproj/project.pbxproj
+++ b/PiwikTracker.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		1F092C141E224C3E00394B30 /* PiwikUserDefaults.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F092C131E224C3E00394B30 /* PiwikUserDefaults.swift */; };
 		1F092C161E225E0200394B30 /* Event.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F092C151E225E0200394B30 /* Event.swift */; };
+		1F092C1A1E26B44500394B30 /* Dispatcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F092C191E26B44500394B30 /* Dispatcher.swift */; };
 		1F1949F21E17A91100458199 /* MemoryQueue.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F1949F11E17A91100458199 /* MemoryQueue.swift */; };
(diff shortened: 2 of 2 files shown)


PiwikTracker/Dispatcher.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+internal protocol Dispatcher {
+    associatedtype Dispatchable
+    
+    var userAgent: String? { get set }
```
minitrue22: Maybe the protocol should only expose the userAgent as { get }
It is up the implementation how the value is create i.e.
* It uses the default value provided by the platform
* It receives the value as a parameter in the init call
But normally it should not be changed after its been assigned

brototyp: True. That is a good point.

PiwikTracker/Dispatcher.swift
```diff
@@ -0,0 +1,11 @@
...
+    
+    func send(event: Dispatchable, success: ()->(), failure: (_ shouldContinue: Bool)->());
```
minitrue22: In practice, do you think `Dispatchable` will be an `Event` or something else?

I think this this protocol to be really useful `Dispatchable` needs to be constraint by something (not sure what it is). If we can not find what this something is maybe just put `Event` there and refactor later.

brototyp: Yes. In our case the Dispatchable is an event. This protocol just defines a Dispatcher as a thing that can dispatch one type (the Dispatchable) in a certain way. So there could be a NSURLSessionEventDispatcher, an AFNetworkingEventDispatcher or even a NSURLSessionSomethingCompletelyDifferentEventDispatcher. I guess in our case it doesn't matter since we only will dispatch Events, hm?

brototyp:
> How should we handle Events that will never be transmitted. For example Events that are to old. It is not possible to submit events that are older than 24h. Should there be an option for the dispatcher to mark an event as We will never be able to transmit this event, don't retry?.

What do you both think about this question? I am still unsure how to cope with this.

PiwikTracker/Dispatcher.swift
minitrue22: I think the Dispatcher protocol is good. I just think we should either constraint the Dispatchable or use Event directly.
You could constraint Dispatchable with a protocol containing one property 'URLRequest' that return the same type. Then let Event implement this protocol.

minitrue22:
There are a couple of strategies.
- could have a max resend count, then simply discard the Event
- a back off strategy for single events or the full dispatcher. If requests fail do not retry within x min
- you can prune old events before sending
- a combination of the above

I am not sure the client needs to prune. Sent the Events and let the server do it.

I would probably defer this to later. But a good solution would be to keep track of the number of send attempts and time stamp for last attempt for events. This can be done inside the Dispatcher.

brototyp:
Oh. Another issue here is sending events in the right order. If I remember correctly events always should be transmitted in the same order they were generated.

Regardless of the order of events in the transmission I like the Idea of retry counts. This way one could discard events after x retries. I am not sure if the Dispatcher is the right place to count retries for an event. I guess we defer this to later and continue for now.

PiwikTracker/Dispatcher.swift
brototyp: I think constraining the Dispatchable to a protocol that serves a URLRequest is not really what we want. In the end the Event should not define its own url. Especially if we keep in mind the second method that dispatches multiple events at once via just one URLRequest. I will refactor the Dispatchable to use Event directly.

PR #120, opened 2017-01-14
WIP: added basic Tracker

brototyp:
Ok, to be honest, I was in quite a struggle here. Some of the older decisions were changed but I am not fully sure about them. I will comment in the code at those places.

Things to do before merging:

- [x] Discuss the open questions
- [x] Add Specs (Best way would be to add a SpecDispatcher and a SpecQueue which expose if they were called and such).


PiwikTracker/Queue.swift
```diff
@@ -1,7 +1,6 @@
 import Foundation
 
-protocol Queue {
-    associatedtype T
+public protocol Queue {
```
brototyp: The Protocol with Associated Type Queue had the downside of adding a Generic to the Tracker.
```
class Tracker<Q:Queue> {
   let queue: Q
}
```

I did not like it this way. It doesn't describe the Tracker properly. Also in the current state it is not necessary. We only want to queue and dispatch Events for now. So I changed the queue to only work with Events (at least for now).

PiwikTracker/Tracker.swift
```diff
@@ -0,0 +1,67 @@
...
+        self.siteId = siteId
+        self.queue = queue
+        self.dispatcher = dispatcher
+        super.init()
+        startDispatchTimer()
+    }
+    
+    private func queue(event: Event) {
+        queue.queue(item: event) { 
+            // Discussion: Do we want the possibility to dispatch synchronous?
```
brototyp: The old SDK hat an option to enable synchronous or rather direct dispatching. So once an event was added to the queue it was immediately dispatched. I do not really to add that feature again. It adds some complexity to the whole code and mostly the same effect can be accomplished by setting the dispatch interval as low as preferred.

If we remove the possibility to transmit Events directly, I would like to remove the Dispatcher.event that transmits only a single event and always use the [BulkTracking api](http://developer.piwik.org/api-reference/tracking-api#bulk-tracking).

PiwikTracker/Tracker.swift
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+    
+    // MARK: dispatching
+    
+    private let dispatchCount = 20
```
brototyp: This is the "How many Events should be dispatched at once". I am still searching for a better name. Any recommendations?

PiwikTracker/Tracker.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+    
+    private func dispatchBatch() {
+        queue.dequeue(withLimit: dispatchCount) { events in
+            self.dispatcher.send(events: events, success: {
+                dispatchBatch()
+            }, failure: { shouldContinue in
+                // Should the events be dispatched in exact order? If so, in case of a failure we cannot continue with the next items.
+                
+                // Also: Dequeueing items an re-queueing items in case of an error is error-prone. If the app crashes during dispatching all currently dispatching events would be lost. And requeueing results in wrong ordered events if we don't sort them otherwise.
```
brototyp: I remember that i read something about transmitting events in exactly the same oder as they were recorded in order to keep the order when analyzing in the backend. Is that true? Is it possible to explicitly define the order by setting the hour, minute and second when transmitting and so adding the possibility of transmitting them in (theoretically) random order? 

@mattab Can you possibly answer that?

brototyp: Also a question to transmitting events. What possible server errors are there and how should there be handled? Is "Try to transmit the Event at a later time." the answer to all errors? Or are there errors that should lead to discarding the Event on the sdk side? For example when transmitting Events that are invalid. Possibly because of they are to old.

Also: When transmitting multiple events at once. Should those events than be handled differently?

@mattab Can you possibly answer that?

PiwikTracker/Tracker.swift
```diff
@@ -0,0 +1,67 @@
...
+                    self.queue.queue(item: event) {}
+                }
+                self.isDispatching = false
+            })
+        }
+    }
+    
+    // MARK: dispatch timer
+    
+    private let dispatchInterval: TimeInterval = 60.0 // Discussion: move this into a configuration?
```
brototyp: I guess exposing as less as possible at first and adding options later if needed is a good way, right?

PiwikTracker/Tracker.swift
mattab: yes @brototyp it's important to send the events, for a given user/visitor, in the right timestamp order. Need to sort events by timestamp before sending them.

mattab: I'd suggest to send the data and as long as the http requests successfully makes it to the Piwik server (ie. no network error or so), the request can be considered successful

mattab: > Also: When transmitting multiple events at once. Should those events than be handled differently?

You can either send them one by one with like 1 second difference, or you could construct a bulk request to send them all at once in one http request. Bulk request doc: http://developer.piwik.org/api-reference/tracking-api#bulk-tracking

brototyp: Alright.

brototyp: Hi @mattab, thanks for your reply. Still the question is how to handle failing transmissions? Either failed single Event transmissions aswell as failed Bulk transmissions. Do you have any hint on how to handle those?

PiwikTracker/Tracker.swift
mattab: (just reading this while answering other questions) maybe `countEventsDispatchedAtOnce`?

brototyp: Nice idea. Count somehow sounds like we are counting things. Maybe use `size`? Like `batchDispatchSize` or so?

PiwikTracker/Queue.swift
```diff
@@ -1,32 +1,20 @@
 import Foundation
 
-protocol Queue {
-    associatedtype T
-    /// The number of queued items.
-    var itemCount: Int { get }
+public protocol Queue {
```
brototyp: I refactored the queue so taking elements doesn't remove them. This way we don't lose events at an app crash during transmission.

PiwikTracker/Dispatcher.swift
```diff
@@ -1,9 +1,7 @@
...
 
-internal protocol Dispatcher {    
+public protocol Dispatcher {
     var userAgent: String? { get }
-    
-    func send(event: Event, success: ()->(), failure: (_ shouldContinue: Bool)->());
-    
-    func send(events: [Event], success: ()->(), failure: (_ shouldContinue: Bool)->());
+        
+    func send(events: [Event], success: ()->(), failure: ()->());
```
brototyp: After an unsuccessful transmission we aren't allowed to send the other events anyways. So the shouldContinue parameter is nonsense.

PiwikTracker/Tracker.swift
mattab: works well as well

PiwikTracker/Tracker.swift
mattab: The very best solutoin would be to send one bulk request containing all pending tracking requests. If it fails, then re-send it later. 

If you're not using a bulk request and only some events fail, then it's problematic to re-send the failed events, because they would be sent in the wrong order. So using bulk request would solve this issue

PiwikTracker/Tracker.swift
thorstenstark: This is perhaps a bit long but what about `numberOfEventsDispatchedAtOnce` ?

PiwikTracker/Tracker.swift
thorstenstark: 👍

PiwikTracker/Dispatcher.swift
thorstenstark: Sounds legit.

PiwikTracker/Tracker.swift
brototyp: Thats a great idea. I took it. Thanks!

Issue #103, opened 2016-12-19
UIApplication-Extension


brototyp:
An extension to return application specific values such as 

- name
- version
- build version and such

Here the same question as in #102 is valid. Should it really be an extension or rather a class.

PR #121, opened 2017-01-21
#103 added Application struct

thorstenstark:
I added an Application struct which creates a new Application instance with all the properties set. Just like we did with the Device struct.


brototyp:
Nothing to add. Thanks!

PR #122, opened 2017-01-30
implemented URLSessionDispatcher

brototyp:
Alright. This is the first version of the URLSessionDispatcher. I think it is ready now and maybe needs some cleanup.


brototyp:
Cleanup ist done. I also adapted some of the Example Application. The `siteId` and Piwik `baseURL` must be configured in the configuration view first. I just submitted the first view events.

Example/iOS Example/AppDelegate.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  AppDelegate.swift
+//  Example
+//
+//  Created by Cornelius Horstmann on 25.02.17.
+//  Copyright © 2017 Mattias Levin. All rights reserved.
+//
```
thorstenstark: I thought we said to remove those comments?!

Example/iOS Example/MenuViewController.m
```diff
@@ -1,77 +0,0 @@
...
-}
-
-
-- (void)alertView:(UIAlertView *)alertView clickedButtonAtIndex:(NSInteger)buttonIndex {
-  
-  // Turn tracking off
-  if (buttonIndex == 0) {
-    [PiwikTracker sharedInstance].optOut = YES;
-  } else {
-    [PiwikTracker sharedInstance].optOut = NO;
```
thorstenstark: Do we have/need tickets/issues for those things that are currently not supported bei the swift version of piwik?

PiwikTracker.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |spec|
   spec.name         = "PiwikTracker"
-  spec.version      = "3.3.0"
-  spec.summary      = "A Piwik tracker written in Objective-C for iOS and OSX apps."
+  spec.version      = "4.0.0-alpha1"
+  spec.summary      = "A Piwik tracker written in Swift for iOS and OSX apps."
```
thorstenstark: What about watchOS and tvOS? Should piwik support those, too?

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -58,13 +58,14 @@ internal struct PiwikUserDefaults {
...
-    func set(clientId: String) {
-        userDefaults.setValue(clientId, forKey: PiwikUserDefaults.Key.visitorID)
-        userDefaults.synchronize()
+    var clientId: String? {
+        get {
+            return userDefaults.string(forKey: PiwikUserDefaults.Key.visitorID)
+        }
+        set {
+            userDefaults.setValue(newValue, forKey: PiwikUserDefaults.Key.visitorID)
+            userDefaults.synchronize()
```
thorstenstark: Do we really need this synchronize?

PiwikTracker/Tracker.swift
```diff
@@ -63,4 +65,60 @@ final public class Tracker: NSObject {
...
         self.dispatchTimer = Timer.scheduledTimer(timeInterval: dispatchInterval, target: self, selector: #selector(dispatch), userInfo: nil, repeats: false)
     }
+    
+    var visitor = Visitor.current()
+    var session = Session.current()
+}
+
+// shared instance
+extension Tracker {
+    public static var sharedInstance: Tracker? {
```
thorstenstark: The Swift way for naming shared instances is usually `shared`, isn't it?

PiwikTracker/Tracker.swift
```diff
@@ -63,4 +65,60 @@ final public class Tracker: NSObject {
...
+            if quality <= 0.5 {
+                break
+            }
+        }
+        return components.joined(separator: ",")
+    }
+}
+
+extension Tracker {
+    public func track(view: [String]) {
```
thorstenstark: As this could be multiple views, I would call it `track(views: [String])`.

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -0,0 +1,126 @@
...
+        request.httpMethod = "POST"
+        request.httpBody = jsonBody
+        send(request: request, success: success, failure: failure)
+    }
+    
+    private func send(request: URLRequest, success: @escaping ()->(), failure: @escaping (_ shouldContinue: Bool)->()) {
+        let task = session.dataTask(with: request) { data, response, error in
+            // should we check the response?
+            // let dataString = String(data: data!, encoding: String.Encoding.utf8)
+            if error == nil {
```
thorstenstark: Does the response perhaps tell us something we should know/handle?

PiwikTracker/Tracker.swift
```diff
@@ -63,4 +65,60 @@ final public class Tracker: NSObject {
...
+// shared instance
+extension Tracker {
+    public static var sharedInstance: Tracker? {
+        get { return _sharedInstance }
+    }
+    
+    public class func configureSharedInstance(withSiteID siteID: String, baseURL: URL) {
+        let queue = MemoryQueue()
+        let dispatcher = URLSessionDispatcher(baseURL: baseURL)
+        self._sharedInstance = Tracker.init(siteId: siteID, queue: queue, dispatcher: dispatcher)
```
thorstenstark: Somehow this feels kind of strange setting the `_sharedInstance` in a class function... But I don't have a better solution for now.

Example/iOS Example/AppDelegate.swift
brototyp: True. I will do so

Example/iOS Example/MenuViewController.m
brototyp: I have just written a ticket #124

PiwikTracker.podspec
brototyp: I would say we start with iOS and add macOS and tvOS later. I don't think it will be a huge issue though.

PiwikTracker/PiwikUserDefaults.swift
brototyp: Sure. Otherwise we might lose the values we just set. I think we should not risk using the values in case of an early exit (especially an app crash). What do you think @thorstenstark ?

PiwikTracker/Tracker.swift
brototyp: Great comment?

- [ ] change the shared instance property to `shared` instead of `sharedInstance`

PiwikTracker/Tracker.swift
brototyp: I actually is not tracking multiple views. A view can be described by a path/categories. For example videos/cats/15. 

Your question definitely shows an issue. All public methods, especially the Tracker, should be documented.

- [ ] Document public Tracker methods.

PiwikTracker/URLSessionDispatcher.swift
brototyp: As far as I know, the result is only interesting when debugging the sdk. The information it returns possibly helps finding sdk-issues.

What do you think about having a debug property in the Tracker itself which is false per default? If it is set, we parse the data and will print it to the console.

PR #128, opened 2017-03-25
Feature/event tracking

brototyp:


PiwikTracker/Tracker.swift
```diff
@@ -143,7 +143,31 @@ extension Tracker {
             actionName: action,
```
thorstenstark: `actionName` is supposed to be a `[String]`?

PiwikTracker/Tracker.swift
```diff
@@ -143,7 +143,31 @@ extension Tracker {
...
             isNewSession: false, // set this to true once we can start a new session
-            referer: nil)
+            referer: nil,
+            eventCategory: nil,
+            eventAction: nil,
+            eventName: nil,
+            eventValue: nil
+        )
+    }
+    internal func event(withCategory category: String, action: String, name: String? = nil, value: Float? = nil) -> Event {
```
thorstenstark: Here the `action` parameter is used for `eventAction`, but in `event(action: [String])` it is used for actionName. I find this irritating. What about using `action` for `actionName` and `eventAction` for `eventAction` consistantly?

brototyp: I don't think so. This way we are hiding complexity from the user. It's all about the context. In this method it's clear that this is an action for an event. The user of this method doesn't have to know that we save the value in the same structure under the hood. And the user doesn't have to know what really an actionName is.

PR #131, opened 2017-04-20
change performBlock ->performBlockAndWait in func deleteEventsWithIDs

YYDD:
I found crashs located in the function "- (void)deleteEventsWithIDs:(NSArray*)entityIDs".
Because the api "- (void)performBlock:(void (^)())block" is asynchronously.Maybe you fecth the entitys from the database when deleting , you will fetch the data is deleting.So i suggest use the api "- (void)performBlockAndWait:(void (^)())block" instand of "- (void)performBlock:(void (^)())block".


brototyp:
Hi @YYDD, thank you for your Pull Request. We dropped ongoing development for version 3 of this SDK. Please consider looking into version 4, which is completely rewritten in Swift.

PR #132, opened 2017-04-30
Custom Dimension - Visit Scope

brototyp:
For now one can only set dimensions on the tracker. They will be used for every event. This is best used for the Visit Scope.


PiwikTracker/Device.swift
```diff
@@ -1,9 +1,9 @@
 import UIKit
 
-internal struct Device {
+public struct Device {
```
brototyp: Not totally sure about these changes here. I like the idea that the user can select the slots for the custom dimensions themselves and can select the values they want to store.

At the same time I am not sure if I want to make Device and Application public.

thorstenstark: I agree that the users should select by themselves which dimensions they want to set unless that doesn't conflict with the piwik API.

I am also not sure about making Device and Application public. It is no feature one would expect from a tracker but I son't see anything serious to be said against it.

PiwikTracker/Tracker.swift
```diff
@@ -186,3 +189,30 @@ extension Tracker {
...
+        remove(dimensionAt: dimension.index)
+        dimensions.append(dimension)
+    }
+    
+    /// Removes a previously set custom dimension.
+    ///
+    /// Use this method to remove a dimension that was set using the `set(value: String, forDimension index: Int)` method.
+    ///
+    /// - Parameter index: The index of the dimension.
+    public func remove(dimensionAt index: Int) {
```
thorstenstark: I would rather call this method just `remove(dimension: CustomDimension)` or `remove(dimensionWith index: Int)`. `dimensionAt` sound like the dimension at the given index in an array should be removed.

brototyp: I see your point. I did it this way to not let the user have to create a CustomDimension object, or even bother about something like Custom dimensions. I didn't want to expose to many types. 
Do you have another idea how to name it?

thorstenstark: Ah okay. As the `CustomDimension` isn't visible to the user/developer I understand your decision. Anyway I would prefer `remove(dimensionWith index: Int)` over `remove(dimensionAt index: Int)` as this rather is, what is happening inside.

PiwikTracker/Device.swift
brototyp: I decided to remove the Device and Application struct from the SDK and moved it to the iOS example application. I think we should stick to our core tasks.

PiwikTracker/Tracker.swift
brototyp: Just changed this. Thanks for the input!

Issue #136, opened 2017-05-07
Starting Sessions


brototyp:
Right now there is now way to start a new session. I am unsure what the best API is to start a new session either on application start/appearance or manually.

API Proposal:

```Swift
Tracker.shared?.newSession()
```

Additionally every Tracker starts a new session on initialization and on enter foreground(observes the `UIApplicationWillEnterForeground` notification).

PR #141, opened 2017-05-19
Feature/136 starting sessions

brototyp:
This PR implements the #136. I slightly changed the API from `newSession()` to `startNewSession()` to be more clear about what this does. Also I thought about when to automatically start a new session. I think the minimal setup for a new Session is on app start. So I did not implement it for app foregrounding.

### DIFF
--- a/PiwikTracker/Tracker.swift
+++ b/PiwikTracker/Tracker.swift
@@ -24,6 +24,7 @@ final public class Tracker: NSObject {
         self.queue = queue
         self.dispatcher = dispatcher
         super.init()
+        startNewSession()
         startDispatchTimer()
     }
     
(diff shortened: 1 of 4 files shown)


PiwikTracker/Tracker.swift
```diff
@@ -24,6 +24,7 @@ final public class Tracker: NSObject {
         self.queue = queue
         self.dispatcher = dispatcher
         super.init()
+        startNewSession()
         startDispatchTimer()
```
brototyp: I wouldn't call `startDispatchTimer()` directly, rather `dispatch()`. `dispatch()` would dispatch the currently stored events and starts a new timer.

In this case I don't think doing any of those is what we want. Starting a new session is not necessary coupled to dispatching events. What do you think?

thorstenstark: I'm fine with that.

PR #144, opened 2017-05-24
Fixed issue where tracking an event wasn’t possible from Objective-C code.

### LINKED ISSUE #142: can't use event Click on oc
can't use event Click on oc.........i can't  find   a  SEL  in  Tracker  about  event

brototyp:
Fixes #142

I also added the ObjectiveCCompatibilityChecker in the test app to ensure that we will keep the Objective-C compatibility.

### DIFF
--- a/PiwikTracker/Tracker.swift
+++ b/PiwikTracker/Tracker.swift
@@ -186,3 +186,13 @@ extension Tracker {
         queue(event: event(withCategory: category, action: action, name: name, value: value))
     }
 }
+
+// Objective-c compatibility extension
+extension Tracker {
+    
+    @objc public func track(eventWithCategory category: String, action: String, name: String? = nil, number: NSNumber? = nil) {
+        let value = number == nil ? nil : number!.floatValue
+        track(eventWithCategory: category, action: action, name: name, value: value)
+    }
+}
+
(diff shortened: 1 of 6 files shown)


PiwikTracker/Tracker.swift
```diff
@@ -186,3 +186,13 @@ extension Tracker {
         queue(event: event(withCategory: category, action: action, name: name, value: value))
     }
 }
+
+// Objective-c compatibility extension
+extension Tracker {
+    
+    @objc public func track(eventWithCategory category: String, action: String, name: String? = nil, number: NSNumber? = nil) {
```
brototyp: This here is the real fix. It is not possible to call a swift method from Objective-C when one parameter is an optional primitive.

Example/iOS Example/ObjectiveCCompatibilityCheker.h
```diff
@@ -0,0 +1,7 @@
+// We use this file to check if all calls are usable from Objective-C code.
+
+#import <Foundation/Foundation.h>
+
+@interface ObjectiveCCompatibilityCheker : NSObject
```
thorstenstark: `Checker`with `ck` ;)

brototyp: Yep, correct 😊
Just fixed that.

Issue #124, opened 2017-03-04
Add the possibility to opt out


brototyp:
My thoughts on the task:

* possibly just a *isEnabled* (or such) property on the tracker
* when opted out every events should be discarded straight away

PR #152, opened 2017-05-29
Feature/124 opt out

brototyp:
Implemented #124


PiwikTracker/PiwikTracker.swift
```diff
@@ -43,6 +54,7 @@ final public class PiwikTracker: NSObject {
     }
     
     internal func queue(event: Event) {
+        guard !isOptedOut else { return }
```
thorstenstark: What happens with events that are already queued?

brototyp: They will be dispatched at the next possibility.

Issue #153, opened 2017-05-29
Setting a custom tracking api parameter


mattab:
Sometimes, new Piwik core releases or even any new plugin may add new parameters to the Tracking API. In those cases it is often not possible to make changes to all necessary SDKs, so the idea is to let user customise the tracking URL. 

-> It would be great to have such a method in 4.0.0 beta :+1: 

Fyi in the PHP SDK we have implemented the following:

```php

    /**
     * Sets a custom tracking parameter. This is useful if you need to send any tracking parameters for a 3rd party
     * plugin that is not shipped with Piwik itself. Please note that custom parameters are cleared after each
     * tracking request.
     *
     * @param string $trackingApiParameter The name of the tracking API parameter, eg 'dimension1'
     * @param string $value Tracking parameter value that shall be sent for this tracking parameter.
     * @return $this
     * @throws Exception
     */
    public function setCustomTrackingParameter($trackingApiParameter, $value)
    {
        $this->customParameters[$trackingApiParameter] = $value;
        return $this;
    }

```
For example users may want to override the Geo location info of country, region and city of a visitor (see https://github.com/piwik/piwik/issues/11563#issuecomment-304547116 for more info how to achieve this without having to set the token_auth)

refs https://github.com/piwik/piwik-sdk-ios/issues/82#issuecomment-161749201

brototyp:
I think the best idea would be a combination of the following two features:

* Add a `let customTrackingParameters: [String:String]` to the Event object
* Make the `queue(event: )` of the `PiwikTracker` public

This way one could create a custom `Event` object, setting all existing parameters. The `customTrackingParameters` dictionary would then enable setting all the different parameters.

mattab:
Is your `Event` related to Piwik Events? http://piwik.org/docs/event-tracking/

maybe it's possible to make it simpler and just have one method like in the PHP tracker, that will set the custom tracking parameter. Similar to 'setResolution', 'setUserId' etc

brototyp:
Events in this SDK are rather "things that should be tracked". So an event could be a view, a download, an exception and so on. Those events are quite atomically. There is barely a situation where one would set any data on the Tracker class directly. In the iOS SDK you create an event and dispatch it via the tracker. There are shortcuts on the Tracker, like `track(view: )` but internally they just create an Event and then dispatch it via itself.

mattab:
[...]

PR #154, opened 2017-06-03
WIP: Added a way of setting customTrackingParameters to an Event and tracking custom Events.

brototyp:
Implements: #153 

* Added a `event()` function to the PiwikTracker that returns an empty event that has core properties already set, such as siteId, visitor, session and such.
* Added a `let customTrackingParameters: [String:String]` property to the Event
* Added a convenience initializer to the Event that copies over all properties from another event and overrides some.
* Added a function to the `PiwikTracker` that tracks a custom `Event`

See all those things in action over [here](https://github.com/piwik/piwik-sdk-ios/pull/154/files#diff-7eb09fef8dbfc0c57ff665277ab105c4R19).

Todo:

- [ ] Discuss if this is a good solution
- [ ] Add Documentation to the new public functions
- [ ] Add an entry in the Changelog
- [ ] Add documentation to the Readme


luohui8891:
Hmm, great, I got it.
We setup `Piwik` in our platform and I should integrate this iOS SDK in our iOS APP.
I found that we just could use only the shared instance with version 3.3 (Objective-C version).
Then I reviewed and upgraded it to version 4.4 (Swift 3), I could create multi-Tracker for different sites (our business needed). But 4.4 didn't support custom variables.

Anyway, I got this branch. Great job.

thorstenstark:
Adding custom parameters sounds like a good idea. But what happens when the custom parameters interfere with some (new) piwik parameters? Should piwik-owned parameters be protected in some kind from being overwritten?

brototyp:
@thorstenstark That's an interesting question. I think it is fine, that the user is able to override existing properties. But we should say so in the documentation.

brototyp:
An alternative implementation got merged.

PR #155, opened 2017-06-05
Feature/macos and tv os

### LINKED ISSUE #134: macOS and tvOS support
I am not sure about the current state of macOS and tvOS support.

- Configure the Framework to be a universal framework
- Configure the Rakefile to run the tests for the macOS and tvOS platforms as well

brototyp:
Closes #134

- Restructured the File Layout: There now is an *Example* folder with folders for all example applications.
- Added tvOS and macOS to the Podspec.
- Setting the Device struct to iOS only.
- Added some platform checks in the URLSessionDispatcher
- Added basic Build tests for the tvOS and macOS platforms


brototyp:
I will fix the conflicts once after we discussed here since it is a quite complex one.

Issue #147, opened 2017-05-25
Add a debug Option to the Tracker.


brototyp:
Right now it is pretty hard to debug the tracker if something goes wrong or not as expected. I think we should be more verbose at some points without being to talky.

- What are events that we want to log per default? 
  - Failed dispatching?
- How should we categorize logs? Just debug yes/no? Or more complex verbose/debug/info/warning/error? I think debug yes/no should be enough.
- How should we log the messages? Always to the console?

brototyp:
How about defining a simple logger protocol like so:

```Swift
protocol Logger {
    func log(message: String)
    func debug(message: String)
}

final class ConsoleLogger: Logger {
    func log(message: String) {
        print("[PiwikTracker] | \(message)")
    }
    func debug(message: String) {
        print("[PiwikTracker] Debug | \(message)")
    }
}

let logger: Logger = ConsoleLogger()
logger.log(message: "This is an important message!")
logger.debug(message: "This is just a debug message.")
```

For now we just use the ConsoleLogger. In the future we can then give users the option to create a custom logger that passes the messages to their own logging mechanism.

PR #156, opened 2017-06-05
Added a Logger that can log messages in different levels.

brototyp:
Closes : #147 

- [x] Add Documentation to the new property
- [x] Add Documentation in the Readme
- [x] Add Documentation on how to write the own Logger
- [x] Add Changelog entry


Example/iOS Example/AppDelegate.swift
```diff
@@ -43,6 +43,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 return
         }
         PiwikTracker.configureSharedInstance(withSiteID: siteid, baseURL: baseurl)
+        PiwikTracker.shared?.logger?.minLevel = .info
```
minitrue22: I know it is not purpose of this PR but,
I would really suggest staying away from singleton and shared instances.

A design goal would be to run multiple trackers instances in the same app.

Let the consumer of the framework create their own shared instance if they want.

minitrue22: Maybe create an issue for tracking this refactoring if we think it is a good idea.

PiwikTracker/Logger.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+public enum LogLevel: Int {
```
minitrue22: Does it have to be public?

PiwikTracker/Logger.swift
```diff
@@ -0,0 +1,55 @@
...
+        case .error: return "E"
+        case .warning: return "W"
+        case .info: return "I"
+        case .debug: return "D"
+        case .verbose: return "V"
+        }
+    }
+}
+
+public protocol Logger {
```
minitrue22: Does it have to be public?

PiwikTracker/Logger.swift
```diff
@@ -0,0 +1,55 @@
...
+    var minLevel: LogLevel { get set }
+    func log(message: ()->(String), with level: LogLevel)
```
minitrue22: Could capture file, function line number in the logs

````
func log(_ message: Void -> String, with level: LogLevel, file: String = #file, function: String = #function, line: Int = #line )
````

the extension methods below also needs the same params for this to work

File: PiwikTracker/Logger.swift [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+    }
+    func warning(message: ()->(String)) {
+        log(message: message, with: .warning)
+    }
+    func error(message: ()->(String)) {
+        log(message: message, with: .error)
+    }
+}
+
+final class DefaultLogger: Logger {
```
minitrue22: I would consider making this a struct.
But `final class` is almost as good :)

PiwikTracker/PiwikTracker.swift
```diff
@@ -22,6 +22,8 @@ final public class PiwikTracker: NSObject {
     private var queue: Queue
     internal let siteId: String
     
+    public var logger: Logger? = DefaultLogger(minLevel: .warning)
```
minitrue22: I would probably make this non optional and not expose the Logger implementations outside the framework.

If the consumer of the framework needs to change the logging level I would hide that behind a method on the PiwikTracker class, e.g. enableDebugLogging that creates and assign new immutable DefaultLogger instance with an appropriate log level.

PiwikTracker/PiwikTracker.swift
```diff
@@ -22,6 +22,8 @@ final public class PiwikTracker: NSObject {
...
+    
     internal static var _sharedInstance: PiwikTracker?
```
minitrue22: I personally think `_` prefixing is a thing of the past but accept if someone else thinks otherwise :)

PiwikTracker/PiwikTracker.swift
```diff
@@ -55,6 +57,7 @@ final public class PiwikTracker: NSObject {
     
     internal func queue(event: Event) {
         guard !isOptedOut else { return }
+        logger?.verbose(){ return "Queued event: \(event)" }
```
minitrue22: Would
````
logger?.verbose { "Queued event: \(event)" }
````
work?

File: Podfile.lock [hunk 9]
```diff
@@ -1,8 +1,8 @@
 PODS:
```
minitrue22: Not sure the .lock file should be tracked by Git.
Maybe add it to .gitignore?

PiwikTracker/PiwikTracker.swift
minitrue22:
````
func verbose(_ message: @autoclosure ()->(String)) { ... }
````
will allow nice syntax
`verbose("123")`

File: Example/iOS Example/AppDelegate.swift (hunk 1 above)
Comment (brototyp) [inline Example/iOS Example/AppDelegate.swift]: There is a ticket (#118) to ensure the Tracker can be instantiated multiple times and I agree that it is necessary to do so. But I don't think we should remove the default, shared logger. I think most users just use a single tracker and are used to have a shared instance. At the same time it doesn't break any other features.

**Comment by brototyp** [comment]
One core feature I wanted to go for in this implementation is a logger that can be replaced. So one could write an adapter for their own Logger (Cocoalumberjack, XCGLogger, ...). That's why the Logger protocol and LogLevel enum as well as there logger property of the Tracker are public.

File: PiwikTracker/Logger.swift (hunk 2 above)
Comment (brototyp) [inline PiwikTracker/Logger.swift]: See the comment below.

File: PiwikTracker/Logger.swift (hunk 3 above)
Comment (brototyp) [inline PiwikTracker/Logger.swift]: See the comment below.

File: PiwikTracker/Logger.swift (hunk 4 above)
Comment (brototyp) [inline PiwikTracker/Logger.swift]: Love it! Good idea!

File: PiwikTracker/Logger.swift (hunk 5 above)
Comment (brototyp) [inline PiwikTracker/Logger.swift]: I rather like to use structs for data structures whose main task is to hold data. For things like this, where the main task is to "do things" or "define behavior" I like to use classes. What do you think about this?

File: PiwikTracker/PiwikTracker.swift (hunk 6 above)
Comment (brototyp) [inline PiwikTracker/PiwikTracker.swift]: See the comment below.

File: PiwikTracker/PiwikTracker.swift (hunk 7 above)
Comment (brototyp) [inline PiwikTracker/PiwikTracker.swift]: Do you think omitting the underscore would be best?

File: PiwikTracker/PiwikTracker.swift (hunk 8 above)
Comment (brototyp) [inline PiwikTracker/PiwikTracker.swift]: Oh! I didn't know about the `@autoclosure`. It's perfect for this use case. Will add it.

File: Podfile.lock (hunk 9 above)
Comment (brototyp) [inline Podfile.lock]: Generally I like to track the .lock file to define which exact versions this is ensured to work with. In this case, since there are only dependencies for the tests, I agree.

File: PiwikTracker/PiwikTracker.swift (hunk 8 above)
Comment (brototyp) [inline PiwikTracker/PiwikTracker.swift]: Just changed this. It's great!

File: PiwikTracker/Logger.swift (hunk 4 above)
Comment (brototyp) [inline PiwikTracker/Logger.swift]: Just changed this. It's great!

File: Example/iOS Example/AppDelegate.swift (hunk 1 above)
Comment (thorstenstark) [inline Example/iOS Example/AppDelegate.swift]: I agree with @brototyp , I also think that users would like to have a shared instance to as a convenient way to use a logger. More ambitious users can use their own instances of loggers.

Comment (minitrue22) [inline Example/iOS Example/AppDelegate.swift]: My experience from looking at a lot of iOS/Mac code is a pattern of overusing singletons. I am just trying to avoid adding to this problem :)

I personally do not see any reason for the framework to provide a singleton. A user that really want to use the singleton pattern in their code can just declare a public variable in the `AppDelegate` or introduce a `Tracker` class with a shared instance or a global variable.

Up to you. I am just on a personal crusade to reduce the use of singletons.

File: PiwikTracker/PiwikTracker.swift (hunk 7 above)
Comment (minitrue22) [inline PiwikTracker/PiwikTracker.swift]: I would not use '_' in any variable or instance names.

Often if you find yourself facing problems with naming resorting to using `_` or other type of prefix/post fix to avoid naming conflicts it is an indication of a design problem.

File: Example/iOS Example/AppDelegate.swift (hunk 1 above)
Comment (minitrue22) [inline Example/iOS Example/AppDelegate.swift]: Maybe look at Alamofire if you want to offer a "default" instance without having to instantiate one.

**Comment by minitrue22** [comment]
> One core feature I wanted to go for in this implementation is a logger that can be replaced. So one could write an adapter for their own Logger (Cocoalumberjack, XCGLogger, ...). That's why the Logger protocol and LogLevel enum as well as there logger property of the Tracker are public.

Ok, I understand, maybe a good idea.
I just think logging is not a core feature for a consumer of the framework. I almost think the framework should never log anything when used in production. Any information that needs to reach the consumer should be a return values or exception. Logging in the context of this framework is only for developers of the framework it self. This will force the developer of the framework to think about the design.

But I understand the purpose and it is ok.

File: PiwikTracker/PiwikTracker.swift (hunk 6 above)
Comment (minitrue22) [inline PiwikTracker/PiwikTracker.swift]: I still suggest making the logger non optional even if exposing it as a property.
The code will become more easy to write and reason if you can avoid optional were possible.

You can provide a default implementation of the ´NoOpLogger´ that user can set to make it easy for them.

File: PiwikTracker/Logger.swift [hunk 10]
```diff
@@ -0,0 +1,56 @@
...
+    var minLevel: LogLevel
+    
+    init(minLevel: LogLevel) {
+        self.minLevel = minLevel
+    }
+    
+    func log(_ message: @autoclosure () -> String, with level: LogLevel, file: String = #file, function: String = #function, line: Int = #line) {
+        guard level.rawValue >= minLevel.rawValue else { return }
+        print("PiwikTracker [\(level.shortcut)] \(message())")
```
minitrue22: Forgot about this. 
`print` is not thread safe (I think).
If this method is called from different thread the print statements will be jumbled up.

You can solve that by doing the printing on a serial queue.
If you do that just must be careful about running `message` on the calling thread before switching to a serial queue or you might get crashes depending on what the provided closure does.

You should probably add a comment on the `Logger` protocol that the implementation needs to deal with any threading themselves. Of instead of exposing the `Logger` to the consumer expose an `Appender` the consumer can set on the `Logger` and let `Logger` handle the threading.

brototyp: Right! Thanks for pointing this out!

PiwikTracker/PiwikTracker.swift
brototyp: I created a `DisabledLogger`.

PR #157, opened 2017-06-08
Screen sizes added to the Event struct 

akshaykolte:
Passed it in the urlsessiondispatcher as well.


PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -92,6 +92,10 @@ fileprivate extension Event {
                 URLQueryItem(name: "m", value: DateFormatter.minuteDateFormatter.string(from: date)),
                 URLQueryItem(name: "s", value: DateFormatter.secondsDateFormatter.string(from: date)),
                 
+                
+                //screen resolution
+                URLQueryItem(name: "res", value: String(describing: screenResolution.width)+"x"+String(describing: screenResolution.height)),
```
brototyp: What do you think about this?

`String(format: "%1.0fx%1.0f", screenResolution.width, screenResolution.height)`

This drops the default `.0` which is precision we don't really need.

akshaykolte:
Can you check why my build fails. It works locally. Also can we merge the code and mark this as fixed?

brototyp:
In the Development branch there now is macOS support. Building the macOS SDK failes. I will copy your branch, fix this issue and merge, once done. Are you fine with that?

akshaykolte:
Yeah. Sounds great. Problem solved.

brototyp:
Was merged in #158

PR #164, opened 2017-07-15
Added a suiteName parameter when instantiating a PiwikTracker to have different stores for different trackers.

brototyp:
I added a new parameter `suiteName` which can be used to distinguish between two separate trackers. This `suiteName` will then be passed through to the UserDefaults in which we are storing the Session and Visitor properties.

What do you think about this idea?

- [x] Add Documentation in the README
- [x] Add a CHANGELOG entry
- [x] Write Migration guide


PiwikTracker/PiwikTracker.swift
```diff
@@ -38,10 +39,13 @@ final public class PiwikTracker: NSObject {
     ///   - siteId: The unique site id generated by the server when a new site was created.
     ///   - queue: The queue to use to store all analytics until it is dispatched to the server.
     ///   - dispatcher: The dispatcher to use to transmit all analytics to the server.
-    required public init(siteId: String, queue: Queue, dispatcher: Dispatcher) {
+    required public init(siteId: String, queue: Queue, dispatcher: Dispatcher, suiteName: String? = nil) {
```
brototyp: Hi @minitrue22, thanks for the feedback. I've been thinking about just using the siteId but there is a downside to it. If one wants to use different Serves and the siteIds on both are different that would lead to a conflict. What do you think about using the siteId and the dispatchers-url in combination for the suiteName? Downside of this would be: What if one changes the URL of the Server or such? That would change all those values.

brototyp: @minitrue22 Do you have any thoughts on that?

mattiaslevin: I agree, use the combination on baseURL and siteId as key., full string or hash, sounds like a good idea.

I would probably not expose it to users of the framework.

PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -5,8 +5,11 @@ import Foundation
 /// All getter and setter are sideeffect free and automatically syncronize
 /// after writing.
 internal struct PiwikUserDefaults {
-    static var standard = PiwikUserDefaults()
-    let userDefaults = UserDefaults.standard
+    let userDefaults: UserDefaults
+
+    init(suiteName: String?) {
```
mattiaslevin: Maybe non optional?

I always to to use non-optimal lets unless there is a strong reason.

mattiaslevin: Your properties could then return non-optional `Date` and `ClientId` by setting a default value when user defaults return nil. This will make the caller code more easier to write without any `guards` and `if let`

Could consider reaming the struct to hide the fact that it is backed by `NSUserDefaults` implementation.

mattiaslevin: I guess for some (advanced) integration tests your need to control the starting values of some of the settings. It might be worth using an interface.

````
protocol SiteSettings {
   var totalNumberOfVisits: Int { get set }
}

extension UserDefaults: SiteSettings {

    func siteSettings(key: String) -> SiteSettings? {
          return UserDefaults(suiteName: key)
   }

    var totalNumberOfVisits: Int {
        get {
            //
        }
        set {
            //
        }
    }
}
````

PiwikTracker/PiwikTracker.swift
brototyp: Yes, I think I will do that. I will add a short comment that the users properties are strongly bound to the server url and site id, so whenever they are lost the settings are lost as well.

PiwikTracker/PiwikUserDefaults.swift
brototyp: > Maybe non optional?
We cannot really make it non optional without loosing compatibility with current installations. They are already using it with suiteName nil

brototyp:
> Your properties could then return non-optional Date and ClientId by setting a default value when user defaults return nil.

At the same time it does make sense that there is no previousVisit stored if it is the first visit. I am not sure if a store should return values that weren't stored before. That is something I wouldn't expect. Especially if the "default" value is so loosely defined and hard to have expectation to.

PiwikTracker/PiwikTracker.swift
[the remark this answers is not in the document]
brototyp: Hey @mattiaslevin, I just found another issue with this. If I would add this right away, we would add a suiteName to Trackers that currently don't have a suite name set. This would result in loosing their data.
So either we port the data over or we change the idea so the "old implementations" would keep their "no suite name" setting. I currently feel like migrating it over would be the best solution.

brototyp:
Ok, I think i figured it out. Every tracker will now store settings in the UserDefault under a suite name. I added a function to migrate over all data from the UserDefaults without a suite name. This migration will be performed, if the shared tracker is configured. This way the tracker can be used as before and no data is lost.

There is currently an edge case, where if one is using the Tracker only without the singleton and then changes to using the singleton, all stored data will be overridden by nil values. That is a case I will prevent.

Beside of it, I am not sure if this is the best version. 

Pros:
- Clean code

Cons:
- Custom migration code
- If a developer is using multiple tracked instances with the same siteId and baseUrl they will override their data (since the store is derived from siteId and baseUrl)

brototyp:
I updated this code, removed the shared instance, and added a manual migration step instead. The downside of this solution is that one might lose their users session data if the migration step is forgotten. But still, I think this is a step in the right direction.

PR #168, opened 2017-07-19
Now can set user agent

luohui8891:


PiwikTracker/Dispatcher.swift
```diff
@@ -4,7 +4,9 @@ public protocol Dispatcher {
     
     var baseURL: URL { get }
     
-    var userAgent: String? { get }
+    var userAgent: String? { get set }
+    
+    func setUserAgent(ua: String)
```
brototyp: It is not necessary to define a specific setter function here, right? I think it is sufficient to use the setter of the property.

PiwikTracker/PiwikTracker.swift
```diff
@@ -128,6 +128,14 @@ final public class PiwikTracker: NSObject {
     internal var visitor = Visitor.current()
     internal var session = Session.current()
     internal var nextEventStartsANewSession = true
+    
+    public func userAgent () -> String {
+        return self.dispatcher.userAgent!
+    }
```
brototyp: I don't think force-unwrapping is a good idea here. The userAgent of a Dispatcher is optional. So it could be that it is nil. In this case it would lead to a crash.

PiwikTracker/PiwikTracker.swift
```diff
@@ -128,6 +128,14 @@ final public class PiwikTracker: NSObject {
...
+    public func setUserAgent (userAgent: String)
+    {
+        self.dispatcher.setUserAgent(ua: userAgent)
+    }
```
brototyp: Better use the property here (as written above).

```Swift
self.dispatcher.userAgent = userAgent
```

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -25,6 +25,10 @@ final class URLSessionDispatcher: Dispatcher {
         return currentUserAgent.appending(" PiwikTracker SDK URLSessionDispatcher")
     }()
     
+    func setUserAgent(ua: String) {
+        self.userAgent = ua
+    }
```
brototyp: If you remove the setter function from the protocol, you could also remove this function here.

brototyp:
Hi @luohui8891, thanks for the Pull Request. Highly appreciated!

I added some comments to the code. In addition to them, can you please:
- [ ] Add some short documentation in the README.md under "Advanced"
- [ ] Add a short documentation for the two new functions in the PiwikTracker.swift
- [ ] Add an entry in the CHANGELOG.md as a feature. Feel free to add a  (by @luohui8891) in there.

Feel free to ask for help. I don't want to change anything in you PR here, it's your work.

PiwikTracker/Dispatcher.swift
minitrue22: To me it feels like the user agent should be immutable once the dispatcher have been created. This will also leave the protocol unchanged.

Instead of making it mutable I suggest providing it as an optional parameter in the constructor in the protocol implementation.

This will make the code safer and more easy to reason about.

PiwikTracker/PiwikTracker.swift
minitrue22: Maybe a computed property?

PiwikTracker/PiwikTracker.swift
minitrue22: Suggest to pass the user agent and the contractor and removing the possibly to change the user agent once from the tracker api.

If someone really really want to the ability the change the user agent after instantiation I suggest they do that by providing their own dispatcher implementation that support mutating the value.

luohui8891:
Hi @brototyp quick review.
I will update the PR by following your comment soon.

PiwikTracker/Dispatcher.swift
luohui8891: @minitrue22 hmm, the thing is that dispatch will create default use agent by itself (SHOULD run in main thread). I think no need to create the default user agent out of the dispatch. What I want is that append some thing in the default user agent, maybe I should add an append method?

luohui8891: Anyway, @brototyp I have removed the special setter function.

PiwikTracker/PiwikTracker.swift
luohui8891: Updated.

PiwikTracker/URLSessionDispatcher.swift
luohui8891: Removed.

PiwikTracker/PiwikTracker.swift
luohui8891: I will keep the mutable `userAgent` property and also add a optional parameter in the init function.
I don't want to create default user agent everywhere.

PiwikTracker/Dispatcher.swift
minitrue22: I guess most users will never need to change the user agent.

Normally I would try to avoid exposing functionality in a public api (especially protocols) that is almost never used. Once you exposed it you need to test it and maintain it forever. 

Then it is better to allow those special users to extend the code or inject their own implementations to meet their goals. One such way is to allow the dispatcher to be instantiated with a customised user agent or maybe create a specialised dispatcher implementation (that inherits or wrap) the default implementation and offers a set or append method. This is still possible without exposing this functionality in the protocol.

luohui8891: @minitrue22 make sense. I have removed the set method.

brototyp:
Hi @luohui8891, thanks for all the effort you put into this PR. It looks really great! 

I have a few short things: 

- [ ] Can you document the newly added parameters in the code? Especially PiwikTracker.swift line 58 and 163 as well as URLSessionDispatcher line 28
- [ ] The tests fail because the ObjectiveCCompatibilityChecker.m calls a nonexistent function
- Are you fine if I tweak the style in a few points?

luohui8891:
@brototyp oh sorry, added the document and updated the `ObjectiveCCompatibilityChecker`.

brototyp:
Thank you @luohui8891, I just merged this. I will be part of the next release.

luohui8891:
@brototyp cool, thanks.
Then I can switch the repo. to official one in my Podfile.

Issue #167, opened 2017-07-19
Dispatch of Events/View only happens Once


jmoukel:
Hey guys,

It seems like the dispatch of events/views only happens once, and not every 30 seconds.

This is the **dispatch** function: 
```
public func dispatch()
{
    guard !isDispatching else {
            logger.verbose("PiwikTracker is already dispatching.")
            return
    }
    guard queue.eventCount > 0 else {
            logger.info("No need to dispatch. Dispatch queue is empty.")
            startDispatchTimer()
            return
    }
    logger.info("Start dispatching events")
    isDispatching = true
    dispatchBatch()
}
```

In the last line, the **dispatchBatch()** function gets called. The **dispatchBatch()** function just sends all the events/views in the queue and then do not call the **startDispatchTimer()** at all to schedule a new execution of the **dispatch()** function. So the subsequent events/views reported never get sent to the server.

brototyp:
Hi @jmoukel, thank you for this awesome bug report and thanks for the effort you put into it. This is an ugly bug. I will fix it right away and will write a test that catches it.

jmoukel:
No problem, thank **you**.

brototyp:
Hi @jmoukel, i just created a Pull Request. Would you review it? It is #169

jmoukel:
@brototyp not allowed do that man, sorry.

brototyp:
Released in version 4.1.0

jmoukel:
Hey man @brototyp, the problem persists.

Even though now **self.startDispatchTimer()** is being called inside **dispatchBatch()** when there're no more events to report, it's being called on a background thread. When you fire a timer on a background thread, it may never call its selector if this background thread terminates for some reason. That's exactly what happens to me.

The following is an extract of the **dispatchBatch()** function. The success param of the **self.dispatcher.send(...)** function is being called on a background thread.
```
self.dispatcher.send(events: events, success: {
     self.queue.remove(events: events, completion:  {
          self.logger.info("Dispatched batch of \(events.count) events.")
          self.dispatchBatch()
     })
}
```

brototyp:
Hi @jmoukel, thanks for the followup. I will dig into it this weekend.

brototyp:
Hi @jmoukel, I just had a look and could reproduce the issue. I think the best thing ist to just ensure the timer is dispatched in the main thread. The pull request is open. I will let you know once it is merged.

brototyp:
[...]

PR #169, opened 2017-07-19
Fixed a bug, where the tracker would stop automatic dispatching.

brototyp:


minitrue22:
#167

PiwikTracker/PiwikTracker.swift
```diff
@@ -96,6 +96,7 @@ final public class PiwikTracker: NSObject {
             guard events.count > 0 else {
                 // there are no more events queued, finish dispatching
                 self.isDispatching = false
+                self.startDispatchTimer()
```
minitrue22: `startDispatchTimer` is called from different levels/abstractions in the code.
Eventually it might become hard to make sure all code branches are covered since the call is spread out in the code.

One option would be if `dispatchBatch()` took a closure that is called when it is done. Then the logic can be confined to the `dispatch()` method (I think).

PiwikTracker/PiwikTracker.swift
```diff
@@ -96,6 +96,7 @@ final public class PiwikTracker: NSObject {
             guard events.count > 0 else {
                 // there are no more events queued, finish dispatching
                 self.isDispatching = false
```
minitrue22: Just noticed that `isDispatching` is not thread safe, depending on what thread the caller calls the `dispatch()` method. There is a risk that the timer is not started/stopped as it should.

minitrue22: I would fix that in a separate PR since the current bug is critical

PiwikTracker/PiwikTracker.swift
```diff
@@ -105,15 +106,20 @@ final public class PiwikTracker: NSObject {
                     self.dispatchBatch()
                 })
             }, failure: { shouldContinue in
-                self.logger.warning("Failed dispatching events with error \(shouldContinue)")
```
thorstenstark: Not part of this PR but what is `shouldContinue`? From the name I would say it's a `Bool`, but it is logged in the warning as error?

brototyp: Thanks for that. I just renamed it.

PiwikTracker/PiwikTracker.swift
brototyp: Maybe it is possible to cleanup the whole dispatching process. It looks a bit overly complex for what it does. I will try to create a proposal.

PR #171, opened 2017-07-29
Adding a short paragraph about using the SDK from Objective-C to the README.

### LINKED ISSUE #165: on the README page link clearly to the Objective-C SDK
Hi @brototyp 
We got a user request asking for the objective-C SDK. Since we offer both, it would be great to add a clear link on the README to link to the "old" objective-C version of the SDK. This will give full flexibility to Piwik users. Thanks in advance!

brototyp:
Resolves #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,23 @@ PiwikTracker.shared?.logger = DefaultLogger(minLevel: .error)
 You can also write your own `Logger` and send the logs whereever you want. Just write a new class/struct an let it conform to the `Logger` protocol.
 
 ### Custom User Agent
-You can now use custom user agent of `PiwikTracker`. Add a parameter for custom user agent in the constructor of `PiwikTracker`.
+The `PiwikTracker` will create a default user agent derived from the WKWebView user agent. This is why you should always instantiate and configure the `PiwikTracker` on the main thread.
+You can instantiate the `PiwikTracker` using your own user agent.
 
-**BTW, be notice, `PiwikTracker` will create a default user agent by creating a web view if you didn't pass the custom one. In another words, you should create instance of `PiwikTracker` in the main thread.**
+```
+PiwikTracker.configureSharedInstance(withSiteID: "5", baseURL: URL(string: "http://your.server.org/path-to-piwik/piwik.php")!, userAgent: "Your custom user agent")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,9 +120,23 @@ PiwikTracker.shared?.logger = DefaultLogger(minLevel: .error)
...
+You can instantiate the `PiwikTracker` using your own user agent.
 
-**BTW, be notice, `PiwikTracker` will create a default user agent by creating a web view if you didn't pass the custom one. In another words, you should create instance of `PiwikTracker` in the main thread.**
+```
+PiwikTracker.configureSharedInstance(withSiteID: "5", baseURL: URL(string: "http://your.server.org/path-to-piwik/piwik.php")!, userAgent: "Your custom user agent")
+```
+
+#### Objective-C compatibility
+
+Version 4 of this SDK is written in Swift, but you can use it in your Objective-C project as well. I you don't want to update you can still use the unsupported older [version 3](https://github.com/piwik/piwik-sdk-ios/tree/version-3). Using the Swift SDK from Objective-C should be pretty straight forward.
```
thorstenstark: **If** you don't want to update...

README.md
```diff
@@ -120,9 +120,23 @@ PiwikTracker.shared?.logger = DefaultLogger(minLevel: .error)
 You can also write your own `Logger` and send the logs whereever you want. Just write a new class/struct an let it conform to the `Logger` protocol.
 
 ### Custom User Agent
-You can now use custom user agent of `PiwikTracker`. Add a parameter for custom user agent in the constructor of `PiwikTracker`.
+The `PiwikTracker` will create a default user agent derrived from the WKWebView user agent. This is why you should always instantiate and configure the `PiwikTracker` in the main thread.
```
thorstenstark: * **derived** from the WKWebView...
* ... **on** the main thread.

minitrue22: Do we need to expose this implementation detail to the user?

Would it be better if the framework run the instantiation of the `WKWebView` on the main thread in stead of pushing this to the user?

I think this should be fixed in code, not be documentation.

brototyp: Absolutely! I will write a ticket for it.

PR #173, opened 2017-08-04
Adds support for Carthage

### LINKED ISSUE #74: Add support for dependency management using Carthage
Framework dependency management.

https://github.com/Carthage/Carthage

elitalon:
This attempts to fix #74 by solving a few build errors that happen when building with `carthage --no-skip-current`.

Theoretically, that should allow projects managed with Carthage to use the framework.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Exclude the build directory
 build/*
+DerivedData
  
 # Exclude temp nibs and swap files
 *~.nib
@@ -19,6 +20,8 @@ xcuserdata
 
 # Exclude generate documentation
 Doc/
+
+# Exclude build products from dependency managers
 Pods
-DerivedData
 /Podfile.lock
+Carthage/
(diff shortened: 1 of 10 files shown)


brototyp:
Hi @elitalon, thank you very much for your effort! I am not used to use Carthage. Is there any way we can add a Carthage-Build check to the travis file?

elitalon:
@brototyp I'll see what I can do. I think running `carthage --no-skip-current` would be a first step, since it's [what they recommend as a first diagnostic](https://github.com/Carthage/Carthage/tree/0.24.0#resolve-build-failures).

brototyp:
@elitalon Your solution might result in wrong test fails. If the `trackerFixture.tracker.queue` happens two times in between dispatches there will only be 4 dispatches in total and the test fails. I will fix this issue in a separate test branch.

brototyp:
@elitalon What do you think about #179?

elitalon:
@brototyp I already commented on #179, much better and without branching on iOS versions. As soon as it gets merged I'll rebase here.

brototyp:
Great, thanks! #179 is merged.

elitalon:
@brototyp The PR is again ready, but something happened with the Travis build. Do you have any idea? I don't have rights to trigger it again.

brototyp:
@elitalon I just started the build once again (a third time) and it all went through. Thanks!

PR #175, opened 2017-08-06
Create user agent on main queue

minitrue22:
## What is new

* Create user agent on main queue
* Refactor duplicate code into common method

## How to test

Run the unit tests
Start and run the test application

## Known problems and limitations
`URLSessionDispatcher` is still not 100% thread safe since we are potentially creating and accessing `userAgent` on different threads. 

I think the best solution would be to always create `URLSessionDispatcher` on the main queue, this will also allow `userAgent` to be declared `let`. But this would require synchronisation in the `PiwikTracker` when creating and accessing the dispatcher, this is probably a necessary change anyway since the `PiwikTracker` is not fully thread safe today.

### DIFF
--- a/PiwikTracker/MainThread.swift
+++ b/PiwikTracker/MainThread.swift
@@ -3,7 +3,3 @@ import Foundation
 func assetMainThread(_ file: StaticString = #file, line: UInt = #line) {
     assert(Thread.isMainThread, "\(file):\(line) must run on the main thread!")
 }
-
-public func asyncMain(_ block: @escaping () -> ()) {
-    DispatchQueue.main.async(execute: block)
-}
--- a/PiwikTracker/MainThread.swift
+++ b/PiwikTracker/MainThread.swift
@@ -1,0 +1,9 @@
+import Foundation
+
+func assetMainThread(_ file: StaticString = #file, line: UInt = #line) {
+    assert(Thread.isMainThread, "\(file):\(line) must run on the main thread!")
+}
+
+public func asyncMain(_ block: @escaping () -> ()) {
+    DispatchQueue.main.async(execute: block)
+}
--- a/PiwikTracker/PiwikTracker.swift
+++ b/PiwikTracker/PiwikTracker.swift
@@ -69,6 +69,8 @@ final public class PiwikTracker: NSObject {
     ///   - baseURL: The url of the piwik server. This url has to end in `piwik.php`.
     ///   - userAgent: An optional parameter for custom user agent.
     convenience public init(siteId: String, baseURL: URL, userAgent: String? = nil) {
+        assert(baseURL.absoluteString.hasSuffix("piwik.php"), "The baseURL is expected to end in piwik.php")
+        
         let queue = MemoryQueue()
         let dispatcher = URLSessionDispatcher(baseURL: baseURL, userAgent: userAgent)
         self.init(siteId: siteId, queue: queue, dispatcher: dispatcher)
--- a/PiwikTracker/URLSessionDispatcher.swift
+++ b/PiwikTracker/URLSessionDispatcher.swift
@@ -23,7 +23,7 @@ final class URLSessionDispatcher: Dispatcher {
         self.baseURL = baseURL
         self.timeout = 5
         self.session = URLSession.shared
-        asyncMain {
+        DispatchQueue.main.async {
             self.userAgent = userAgent ?? URLSessionDispatcher.defaultUserAgent()
         }
     }
--- a/PiwikTracker/URLSessionDispatcher.swift
+++ b/PiwikTracker/URLSessionDispatcher.swift
@@ -12,7 +12,24 @@ final class URLSessionDispatcher: Dispatcher {
     let session: URLSession
     let baseURL: URL
 
-    var userAgent: String? = {
+    private(set) var userAgent: String?
+    
+    /// Generate a URLSessionDispatcher instance
+    ///
+    /// - Parameters:
+    ///   - baseURL: The url of the piwik server. This url has to end in `piwik.php`.
+    ///   - userAgent: An optional parameter for custom user agent.
+    init(baseURL: URL, userAgent: String? = nil) {                
+        self.baseURL = baseURL
+        self.timeout = 5
+        self.session = URLSession.shared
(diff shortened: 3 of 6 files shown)


Example/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -352,7 +352,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n";
+			shellScript = "diff \"${PODS_ROOT}/../Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n";
```
minitrue22: Cocoapods did these changes, not sure they are correct or now.

PiwikTracker/PiwikTracker.swift
```diff
@@ -57,6 +57,8 @@ final public class PiwikTracker: NSObject {
     ///   - baseURL: The url of the piwik server. This url has to end in `piwik.php`.
     ///   - userAgent: An optional parameter for custom user agent.
     convenience public init(siteId: String, baseURL: URL, userAgent: String? = nil) {
+        assert(baseURL.absoluteString.hasSuffix("piwik.php"), "The baseURL is expected to end in piwik.php")
```
minitrue22: Moving the assert as close to the call as possible

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -12,46 +12,44 @@ final class URLSessionDispatcher: Dispatcher {
...
-            fatalError("The baseURL is expected to end in piwik.php")
-        }
+    init(baseURL: URL, userAgent: String? = nil) {                
         self.baseURL = baseURL
         self.timeout = 5
         self.session = URLSession.shared
-        if (userAgent != nil) {
-            self.userAgent = userAgent
+        asyncMain {
+            self.userAgent = userAgent ?? URLSessionDispatcher.defaultUserAgent()
```
minitrue22: This this is not 100% thread safe since we might access self.userAgent on something else then the main queue.

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -67,14 +65,19 @@ final class URLSessionDispatcher: Dispatcher {
...
-        var request = URLRequest(url: baseURL, cachePolicy: .reloadIgnoringCacheData, timeoutInterval: timeout)
-        request.setValue("application/json; charset=utf-8", forHTTPHeaderField: "Content-Type")
-        request.setValue(userAgent, forHTTPHeaderField: "User-Agent")
-        request.httpMethod = "POST"
-        request.httpBody = jsonBody
+        let request = buildRequest(baseURL: baseURL, method: "POST", contentType: "application/json; charset=utf-8", body: jsonBody)
         send(request: request, success: success, failure: failure)
     }
     
+    private func buildRequest(baseURL: URL, method: String, contentType: String? = nil, body: Data? = nil) -> URLRequest {
```
minitrue22: Reducing code duplication

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -124,11 +126,11 @@ fileprivate extension Event {
...
                 URLQueryItem(name: "e_v", value: eventValue != nil ? "\(eventValue!)" : nil),
                 
-                ].filter({ $0.value != nil }) // remove the items that lack the value
-            for dimension in dimensions {
-                items.append(URLQueryItem(name: "dimension\(dimension.index)", value: dimension.value))
-            }
-            return items
+                ].filter { $0.value != nil }
+            
+            let dimensionItems = dimensions.map { URLQueryItem(name: "dimension\($0.index)", value: $0.value) }
```
minitrue22: Less imperative

PiwikTracker/MainThread.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+func assetMainThread(_ file: StaticString = #file, line: UInt = #line) {
+    assert(Thread.isMainThread, "\(file):\(line) must run on the main thread!")
+}
+
+public func asyncMain(_ block: @escaping () -> ()) {
+    DispatchQueue.main.async(execute: block)
+}
```
brototyp: I do like the `assertMainThread` function but I'm not totally convinced by the `asyncMain`. In the end it adds complexity and an extra abstraction for just saving 15 characters. Is it really worth it?

PiwikTracker/URLSessionDispatcher.swift
brototyp: I like it.

PiwikTracker/URLSessionDispatcher.swift
brototyp: Nice idea

brototyp:
Thanks @minitrue22 for this PR. I love your detailed PR description!

PiwikTracker/MainThread.swift
brototyp: I've been so presumptuous to remove this function and replace it by directly calling `DispatchQueue.main.async` instead. I hope you are fine with that. Please let me know if not. I am looking forward to merge this soon.

PiwikTracker/URLSessionDispatcher.swift
brototyp: I thinks that's fine so far. There's barely a change that this is false. Even if it will be corrected with the next request. In the end, the User Agent is "just" needed to extract some information about the device, such as the OS. If the first request lacks the data, the next one will add it, which will fill in the blank first spots.

PR #176, opened 2017-08-06
Make "Objective-C compatibility" header bigger

mattab:


brototyp:
What do you think about instead of making the header bigger to link this header from right from the top? Something like "written in Swift, [compatible with Objective-C](link to the header)"?

brototyp:
@mattab I just did that. Let me know if you are fine with this change.

mattab:
:+1: 

 Maybe also rename "Advanced" to "Advanced usage" and make titles below the right level so they are bigger and there are title less level.

brototyp:
Good point.

PR #179, opened 2017-09-04
Added the AutoTracker to the tests to make them iOS 9.3 compatible

brototyp:
To fix a testing issue as described here:  #173


elitalon:
Looks good to me 👍🏼

I would move `AutoTracker` as a private class inside `TrackerSpec.swift`. Just to keep it in context, as it is indeed a helper object.

brototyp:
Good point. Changed just that.

PR #181, opened 2017-09-08
Feature/180 custom visitor ID

niksawtschuk:
Small changes here to add a public variable which allows apps to set the custom user ID for the shared tracker's current visitor object. I used the existing structure that was defined already, so this PR is really just connecting the plumbing and enhancing the docs/example app.


PiwikTracker/PiwikUserDefaults.swift
```diff
@@ -60,10 +60,20 @@ internal struct PiwikUserDefaults {
...
         get {
-            return userDefaults.string(forKey: PiwikUserDefaults.Key.visitorID)
+            return userDefaults.string(forKey: PiwikUserDefaults.Key.clientID)
         }
         set {
-            userDefaults.setValue(newValue, forKey: PiwikUserDefaults.Key.visitorID)
+            userDefaults.setValue(newValue, forKey: PiwikUserDefaults.Key.clientID)
+            userDefaults.synchronize()
+        }
+    }
```
brototyp: If you change it this way, it will change the behavior of all old implementations. For all old Implementations the `clientId` was saved under the key `PiwikUserDefaults.Key.visitorID`. After this change, all existing installations will use the old `clientId` as their new `visitorId`. I thinks that's an issue.

Example/ios/iOS Example/UserIDViewController.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import PiwikTracker
+
+class UserIDViewController: UIViewController {
```
brototyp: Great! Thanks for adding this to the Example Application!

PiwikTracker/PiwikUserDefaults.swift
niksawtschuk: Good point. I've adjusted this in a9e9bf390f54fc714bbb49bbef760ace4d993b6a. I left my changes for using the correct "key name", but changing what value the key is using so it's using the old one. This should retain the existing functionality but be more clear in the future.

brototyp: Yeah, I think that's the best solution in this case. It's a bit of a pity that it got introduced in the first place. Thanks for changing it this way.

brototyp:
Thanks @niksawtschuk, for this PR. I would like to add your GitHub handle to the CHANGELOG.md. Are you fine with that?

niksawtschuk:
No problem @brototyp.

PR #188, opened 2017-10-02
Added a way of setting customTrackingParameters to an Event and tracking custom Events.

brototyp:
This is an implementation of #153, as an alternative of #154 which is much easier.

Example usage:
```swift
    private func example() {
        guard let piwikTracker = PiwikTracker.shared else { return }
        let downloadURL = URL(string: "https://builds.piwik.org/piwik.zip")!
        let event = Event(tracker: piwikTracker, action: ["menu", "some page", "downloadfile"], customTrackingParameters: ["download":downloadURL.absoluteString])
        PiwikTracker.shared?.track(event)
    }
```

Upsides: 
- Much easier usage. Create the event directly and the track it. No need to create intermediary events.
- Adds only two new functions. A new initializer for the Event and a new function to track an event directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+* **feature** Added the ability to send custom events with custom tracking parameters. [#153](https://github.com/piwik/piwik-sdk-ios/issues/153)
+
 ## 4.2.0
 * **feature** Added ability to customize user agent. [#168](https://github.com/piwik/piwik-sdk-ios/pull/168)
 * **feature** Added ability to customize User ID.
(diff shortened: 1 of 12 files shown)


brototyp:
@minitrue22 Do you have any input to this?

thorstenstark:
I like this way of setting custom parameters. But I'm currently not sure if this may have effects on the compatibility with obj-c.

brototyp:
What currently works with objc will still work. We didn't change the interface. I'm not sure if we will be able to pull of objc compatibility for the new features.

brototyp:
@thorstenstark I just implemented the rest, added example code and wrote the documentation.

Issue #111, opened 2016-12-20
Custom Dimension Tracking


brototyp:
- [x] ~Adding Custom Dimensions in the Visit Scope [Pull Request](https://github.com/piwik/piwik-sdk-ios/pull/132)~
- [x] Adding Custom Dimensions for the Action Scope

jmoukel:
Just wondering. When are you guys planing to release this?

brototyp:
Hi @jmoukel, thanks for asking. I just did the final touches on the Pull Request that adds the Custom Dimensions for the Visit Scope. For the 4.1 release only the #118 is left. I think it will be finished later this month. If you can't wait, just try the development branch.

brototyp:
Also: Feedback on the Custom Dimensions implementation is welcome. I am currently not sure how important the Action Scope is.

jmoukel:
Hi @brototyp ,

Sounds good, thanks. I'll wait until you finish it 💯 given it might be ready soon. Good job, btw.

I'm new with Piwik, but it seems the scope is very important for _Custom Dimensions_. I can see myself using the scope "**Action**" in my apps and I will definitely use the scope "**Visit**". I mean, I found very useful to report info both visit-wise (_e.g. app version being used_) and action-wise (_e.g. device volume when Play button was pressed_).

jmoukel:
Hi @brototyp ,

I saw you added Visit scope Custom Dimensions to the main branch. Pretty cool indeed.

_Did you decide to also add Action scope custom dimensions?_ No rush or anything.

jmoukel:
Hi @brototyp !
Just wondering, when are you guys planing to release the **Action Scope** Custom Dimensions?

Thanks :]

brototyp:
Hi @jmoukel, thanks for nudging. Sometimes I need someone to remind me about a certain feature. I focussed on implementing #153 where I think we have a near to final pull request in #188.

Once this PR is through, I will focus on this one here.

jmoukel:
Thank you :] @brototyp

brototyp:
It just got merged and will be part of the next release

brototyp:
Just released [version 4.4.0](https://github.com/piwik/piwik-sdk-ios/releases/tag/v4.4.0) containing this feature.

PR #193, opened 2017-10-09
Feature/111 custom dimensions with action scope

brototyp:
Implements #111 for the action scope.

Basically I made the `CustomDimension` struct public and added an optional parameter to the tracking functions to add additional dimensions.

Since the `CustomDimension` is a struct, it is not usable from Objective-C. Still I think this is the best idea.

### DIFF
--- a/PiwikTracker/PiwikTracker.swift
+++ b/PiwikTracker/PiwikTracker.swift
@@ -209,21 +209,36 @@ extension PiwikTracker {
 }
 
 extension PiwikTracker {
+    
+    /// Tracks a custom Event
+    ///
+    /// - Parameter event: The event that should be tracked.
     public func track(_ event: Event) {
         queue(event: event)
     }
+    
     /// Tracks a screenview.
     ///
     /// This method can be used to track hierarchical screen names, e.g. screen/settings/register. Use this to create a hierarchical and logical grouping of screen views in the Piwik web interface.
     ///
     /// - Parameter view: An array of hierarchical screen names.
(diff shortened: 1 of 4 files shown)


PiwikTracker/PiwikTracker.swift
```diff
@@ -260,13 +216,15 @@ extension PiwikTracker {
     ///
     /// - Parameter view: An array of hierarchical screen names.
     /// - Parameter url: The url of the page that was viewed. If none set the url will be http://example.com appended by the screen segments. Example: http://example.com/players/john-appleseed
```
thorstenstark: The documentation here should mention the `dimensions` parameter.

PiwikTracker/PiwikTracker.swift
```diff
@@ -260,13 +216,15 @@ extension PiwikTracker {
...
-    @objc public func track(view: [String], url: URL? = nil) {
-        queue(event: event(action: view, url: url))
+    public func track(view: [String], url: URL? = nil, dimensions: [CustomDimension] = []) {
```
thorstenstark: As `CustomDimension` isn't visible to objc, I'm not sure if these methods are still visible to objc.

PiwikTracker/PiwikTracker.swift
brototyp: Good catch! Done

PiwikTracker/PiwikTracker.swift
brototyp: There is a Objective-C extension without the dimensions property. See https://github.com/piwik/piwik-sdk-ios/pull/193/files#diff-bd7b3c5478dc6552eb40fd9baaa55d18R289

Issue #194, opened 2017-10-11
Event tracking: remove "example.com" url.


ferenclakos:
Hi!

I would like use the event tracking but have a problem.

We can see the events on the dashboard but we would like remove the "example.com" url from actions.
As I have seen it has been hard coded into source code.

<img width="396" alt="event_tracking_problem" src="https://user-images.githubusercontent.com/4071386/31433856-e7ce0280-ae7a-11e7-9e62-e88d8f79aea7.png">

Could you please provide a possibility to change this url?

Thanks.

rdani91:
We have this similar problem too.
Our system requires to set a specific domain, but we couldn't find the possibility to do that.

thbaja:
You can create your own custom events where you can set the URL as described [here](https://github.com/piwik/piwik-sdk-ios#sending-custom-events).

But it would be nice to be able to set a base URL once, maybe in `configureSharedInstance` (not to be confused with the baseURL tracking endpoint).And then screen names would use this url as the base when tracking page views `(PiwikTracker.shared?.track(view: ["path","to","your","page"]))`

And similar behaviour when tracking events. Ie. the url sent with the event would be this base url, along with optional screen names.

YvesCandel:
Is the URL necessary? I would rather just leave it empty/null... the Android SDK works like that it seems.

brototyp:
I think the easiest way for now would be to add a url parameter to the event tracking function. This would be consistent with tracking a view (where you can already do that).

Additionally I like the idea of defining the "http://example.com" once and we use it to generate the url if none is given.

As described [here](https://developer.piwik.org/api-reference/tracking-api) the url is a required field. @mattab Is this really the case? Is it possible to ditch the url since in most/some cases one on mobile there aren't urls matching the content in the app?

YvesCandel:
We're seeing this from our Android app:

![screen shot 2017-10-12 at 09 42 44](https://user-images.githubusercontent.com/6763192/31484759-d7e45704-af31-11e7-8d43-f0557746e2d8.png)

...and this from our iOS app:

![screen shot 2017-10-12 at 09 42 52](https://user-images.githubusercontent.com/6763192/31484771-df226f4c-af31-11e7-901f-50c5a7ec3755.png)

This somehow makes me doubt that the url is a required field...

mattab:
[...]

PR #195, opened 2017-10-14
Removed the url as required parameter

brototyp:
Fixes/Implements #194

What changed:

- The url parameter of an `Event` is now an optional parameter
- The convenience methods for view and event tracking have a new optional url parameter
- When not defining a url, no default url will be generated

Do we want to add a way to generate a default url? If so, this would be my proposal:

- Add a new property `contentBaseUrl` to the PiwikTracker
- If this property is set, every event that has no url set will generate one by its own
- To generate a url, it will use the `contentBaseUrl` and append the action-array as path

Example:
- `contentBaseUrl` is https://piwik.org
- actions is ["docs", "requirements"]
- the url will be https://piwik.org/docs/requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * **feature** Added a way to add custom dimension in the action scope. [#111](https://github.com/piwik/piwik-sdk-ios/issues/111)
 * **improvement** The PiwikTracker is now save to create in a background thread. [#175](https://github.com/piwik/piwik-sdk-ios/pull/175)
 * **improvement** The Logger is now objc compatible. [#185](https://github.com/piwik/piwik-sdk-ios/issues/185)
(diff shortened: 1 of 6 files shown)


thbaja:
I think this is a good PR.

And a very good proposal regarding the contentBaseUrl. Perhaps this should also apply for when tracking screens?

brototyp:
Yes, I would apply the logic to all events that are tracked. It is a bit complicated to differentiate between an event and an `Event` where the latter includes events, screen views and all other kind of data send to the backend.

thbaja:
Understood 👍

brototyp:
Just released [version 4.4.0](https://github.com/piwik/piwik-sdk-ios/releases/tag/v4.4.0) containing this change.

PR #198, opened 2017-10-29
Fix documentation typos

elitalon:
I fixed a few typos in a couple of Markdown files.


brototyp:
Awesome! Thanks for the effort!

Issue #191, opened 2017-10-09
Give access to the structs Application and Device.


thbaja:
Currently the two structs have `internal` as access level (in _Application.swift_ and _Device.swift_).

But the two structs are quite handy when using Custom Dimensions, exactly as shown in the `piwik-sdk-ios/Example/ios/iOS Example/CustomDimensionsViewController.swift`.

Having them public instead would be very useful.

brototyp:
Hi @thbaja, thanks for noticing this. I think the Device struct could be made public, but the Application struct is currently only ios compatible. Before making it public we should add compatibility for macOS and tvOS. What do you think?

thbaja:
The repository is for an iOS SDK so I don't think it would be unreasonable to make it public without support for macOS tvOS :-) Also given the conditional compilation currently in place.

But it would be nice to have macOS and tvOS support at some point, also now where everything else is compatible. 
We could wait until someone requests it? Perhaps I have some time next week to look at it.

brototyp:
I see your point but, despite the repository name, this tracker supports tvOS and macOS. If you are able to dig into it to find a solution for both it would be great!

brototyp:
@thbaja I just pushed a PR where I made both public. Please have a look, if you can spare some time.

brototyp:
Just released [version 4.4.0](https://github.com/piwik/piwik-sdk-ios/releases/tag/v4.4.0) containing this.

PR #201, opened 2017-11-01
Updated the Application and Device structs to be public

brototyp:
Implements #191

It was pretty straight forward. The Application struct actually works right out of the box and the Device struct was already migrated to tvOS and macOS.

I found a bug, where the documentation of the `bundleName` was suggesting that it returns the bundle identifier. I fixed it and added a `bundleIdentifier` property.


thbaja:
Looks good to me 👍

PR #205, opened 2017-11-23
Feature/custom variables

zantoku:
This Pull Request adds support for Custom Variables, largely following the implementation pattern of Custom Dimensions. Custom Variables can be set scoped to visit (as a property of the shared PiwikTracker) or to an Event. Also, the tracker now generates the same default variables (Platform, OS Version, App Version) as the old 3.x tracker.

Also, I've improved the ObjC-interoperability with Custom Dimensions, by adding a setter method that can be called from there.

### DIFF
--- a/PiwikTracker/PiwikTracker.swift
+++ b/PiwikTracker/PiwikTracker.swift
@@ -141,7 +141,7 @@ final public class PiwikTracker: NSObject {
     
     // MARK: dispatch timer
     
-    public var dispatchInterval: TimeInterval = 30.0 {
+    @objc public var dispatchInterval: TimeInterval = 30.0 {
         didSet {
             startDispatchTimer()
         }
--- a/PiwikTracker/PiwikTracker.swift
+++ b/PiwikTracker/PiwikTracker.swift
@@ -9,7 +9,7 @@ final public class PiwikTracker: NSObject {
     
     /// Defines if the user opted out of tracking. When set to true, every event
     /// will be discarded immediately. This property is persisted between app launches.
-    public var isOptedOut: Bool {
+    @objc public var isOptedOut: Bool {
         get {
             return PiwikUserDefaults.standard.optOut
         }
--- a/PiwikTracker/PiwikTracker.swift
+++ b/PiwikTracker/PiwikTracker.swift
@@ -36,6 +36,8 @@ final public class PiwikTracker: NSObject {
 
     internal var dimensions: [CustomDimension] = []
     
+    @objc public var useDefaultCustomVariables: Bool = false
+    private lazy var cvars: [CustomVariable] = getDefaultCVars()
     
     /// This logger is used to perform logging of all sorts of piwik related information.
     /// Per default it is a `DefaultLogger` with a `minLevel` of `LogLevel.warning`. You can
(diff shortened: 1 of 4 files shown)


PiwikTracker/PiwikTracker.swift
```diff
@@ -281,18 +283,70 @@ extension PiwikTracker {
...
         dimensions = dimensions.filter({ dimension in
             dimension.index != index
         })
     }
 }
 
+
+extension PiwikTracker {
+    
+    internal func getDefaultCVars() -> [CustomVariable] {
```
brototyp: What do you think about moving this to the `CustomVariable` struct? Something like this:
```Swift
extension CustomVariable {
    static var defaultVariables: [CustomVariable] {
        // return ...
    }
}
```

PiwikTracker/PiwikTracker.swift
```diff
@@ -36,6 +36,8 @@ final public class PiwikTracker: NSObject {
 
     internal var dimensions: [CustomDimension] = []
     
+    @objc public var useDefaultCustomVariables: Bool = false
+    private lazy var cvars: [CustomVariable] = getDefaultCVars()
```
brototyp: I don't like the `cvars` variable name here to much. We usually don't use abbreviations. What do you think about using `customVariables` here? This then shows another minor issue. It is imho hard to distinguish (just by the variable names) between the `cvars` and the `customVariables` that you implemented as a slice from these here.

What do you think about removing the `useDefaultCustomVariables` completely and let the users of the SDK add the "default" variables themselves? Exposing a `defaultVariables` property on the `CustomVariable` for example could help here. This way, the code here in the SDK is much cleaner and it is more direct for the user.

PiwikTracker/PiwikTracker.swift
```diff
@@ -281,18 +283,70 @@ extension PiwikTracker {
...
+        cvars.append(CustomVariable(name: name, value: value))
+        return cvars.count
+    }
+    
+    /// Remove a previously set Custom Variable. Note that the default Custom Variables cannot be removed.
+    ///
+    /// Note that, as with any array, the index of any succeeding Custom Variables is decremented by 1.
+    ///
+    /// - Parameter index: The index that was previously returned by addCustomVariable().
+    @objc public func removeCustomVariableAtIndex(_ index: Int) {
```
brototyp: What do you think about referencing the variables by name, not by index? In this case I think the `CustomVariable` struct should bring the `index` to be referenced in the API.

```Swift
@objc public func removeCustomVariable(withName name: String) {
    cvars = cvars.filter{$0.name != name }
}
```

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -94,9 +94,15 @@ final class URLSessionDispatcher: Dispatcher {
...
+    
+    private func cvarParameterValue() -> String {
+        var cvars: Array<String> = customVariables.enumerated().map { "\"\($0.offset + 1)\":[\"\($0.element.name)\",\"\($0.element.value)\"]" }
+        return "{\(cvars.joined(separator: ","))}"
+    }
+    
     var queryItems: [URLQueryItem] {
         get {
-            let items = [
+            var items = [
```
brototyp: Can you keep this as a let and the join al the different query items together in the end?

PiwikTracker/URLSessionDispatcher.swift
```diff
@@ -94,9 +94,15 @@ final class URLSessionDispatcher: Dispatcher {
 }
 
 fileprivate extension Event {
+    
+    private func cvarParameterValue() -> String {
+        var cvars: Array<String> = customVariables.enumerated().map { "\"\($0.offset + 1)\":[\"\($0.element.name)\",\"\($0.element.value)\"]" }
+        return "{\(cvars.joined(separator: ","))}"
```
brototyp: I think this code is a bit to complex to parse. Do you think it is possible to optimize it?

rafalkitta:
I would appreciate if you add this feature.

brototyp:
I will. There is only a few little things left. @zantoku let me know if you don't agree with some of my comments or if you can't find any time for them.

brototyp:
Hi @zantoku, thanks for your effort here. @manuroe improved your PR and we merged it in #223

PR #211, opened 2017-12-14
Fix a couple Swift 4 complaints

nwhittaker:
* Fix warning: "'substring(with:)' is deprecated: Please use String slicing subscript."
* Fix subsequent error: "Subscripts returning String were obsoleted in Swift 4; explicitly construct a String from subscripted result."


brototyp:
Awesome. Thanks!

Issue #208, opened 2017-12-04
Compiler Error


mkrajcir:
I tried to use Piwik iOS SDK 4.4. Unfortunatelly I got a Compiler Error with Xcode 9.1.

`Pods/PiwikTracker/PiwikTracker/Application.swift:4:27: 'bundleDisplayNameForCurrentApplication' is inaccessible due to 'private' protection level`

**Solution:**
`private` methods from extensions should be `fileprivate`.

brototyp:
Hi @mkrajcir, thanks for your report. Unfortunately I wasn't able to reproduce your issue. What exactly can be done to reproduce it? The iOS Example in the repository works well and doesn't show this issue. Thanks in advance!

AndreaGer:
@brototyp I get the same issue. Just install the latest version of the pod and try to compile it. You don't actually need to implement the code to get the issue. 4.4 is useless as of now.

mrommel:
Same for me - I was on 
`pod 'PiwikTracker', '4.2.0'`
and updating to
`pod 'PiwikTracker', '4.4.0'`
gives a lot of build errors:

> /myfunnyapp/Pods/PiwikTracker/PiwikTracker/Application.swift:4:27: error: 'bundleDisplayNameForCurrentApplication' is inaccessible due to 'private' protection level
>         let displayName = bundleDisplayNameForCurrentApplication()
>                           ^
> /myfunnyapp/Pods/PiwikTracker/PiwikTracker/Application.swift:30:25: note: 'bundleDisplayNameForCurrentApplication()' declared here
>     private static func bundleDisplayNameForCurrentApplication() -> String? {
>                         ^
> /myfunnyapp/Pods/PiwikTracker/PiwikTracker/Application.swift:5:20: error: 'bundleNameForCurrentApplication' is inaccessible due to 'private' protection level
>         let name = bundleNameForCurrentApplication()
>                    ^
> /myfunnyapp/Pods/PiwikTracker/PiwikTracker/Application.swift:35:25: note: 'bundleNameForCurrentApplication()' declared here
>     private static func bundleNameForCurrentApplication() -> String? {
>                         ^
> /myfunnyapp/Pods/PiwikTracker/PiwikTracker/Application.swift:6:26: error: 'bundleIdentifierForCurrentApplication' is inaccessible due to 'private' protection level
>         let identifier = bundleIdentifierForCurrentApplication()
>                          ^
[...]

PR #214, opened 2017-12-18
fix compiler error of https://github.com/piwik/piwik-sdk-ios/issues/208

mrommel:
- make application functions fileprivate instead of private


brototyp:
Closing this PR since it appears to be a issue using the SDK. See #208 for more details.

PR #223, opened 2018-01-29
Another PR for Feature/custom variables

manuroe:
Hi,

This PR is the continuation of the work made by @zantoku in PR #205, which is stuck because review remarks have not yet been taken into account.

What I have done in this PR:
- merge the @zantoku's PR into the current develop branch
- take into account @brototyp's remarks except https://github.com/matomo-org/matomo-sdk-ios/pull/205#discussion_r154520261 because I have no idea how to make the string formatting in this piece of code more readable.
- remove default custom variables created by the initial `getDefaultCVars` method because indexes and names used for these custom variables look like very specific to an app. I have not seen  matomo using such default values in other SDKs. So, I have preferred to remove them and let the app set them.

This last point will require to make the `Device` struct accessible from objc in order to offer the same functionality as the initial PR. If you are ok, I will make it in another PR.

### DIFF
--- a/PiwikTracker.xcodeproj/project.pbxproj
+++ b/PiwikTracker.xcodeproj/project.pbxproj
@@ -33,6 +33,7 @@
 		1FCFF0241F82C7A50038BC17 /* CustomDimension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FDC917B1F1A64BB0046F506 /* CustomDimension.swift */; };
 		1FDC917F1F1A65150046F506 /* Application.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FDC917D1F1A65150046F506 /* Application.swift */; };
 		1FDC91801F1A65150046F506 /* Device.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FDC917E1F1A65150046F506 /* Device.swift */; };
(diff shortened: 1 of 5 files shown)


manuroe:
The last point is implemented in https://github.com/matomo-org/matomo-sdk-ios/pull/224.

brototyp:
Hi @manuroe, Awesome. Thank you for your work. To the `getDefaultCVars`: I think the code came from porting over the original, Objective-c code. Back then such a logic was implemented. I am very fine with keeping it as you implemented it and to deprecate such usage.

I will sketch around a bit regarding my comment about the readability.

brototyp:
I thought about serializing it as JSON, but that adds another complexity. The best I could find is the following. What do you think?

```Swift
private func customVariableParameterValue() -> String {
    let customVariableParameterValue: [String] = customVariables.map { "\"\($0.index)\":[\"\($0.name)\",\"\($0.value)\"]" }
    return "{\(customVariableParameterValue.joined(separator: ","))}"
}
```

manuroe:
Your `customVariableParameterValue` is fine by me. I have updated the PR.

brototyp:
Awesome! Thanks.

brototyp:
@manuroe I've just seen, that there is no Changelog entry. Do you want to be named there?

manuroe:
>Do you want to be named there?

@brototyp, yes please.

PR #224, opened 2018-01-29
Make Device.swift usable from objc

manuroe:
Hi,
This PR turns Device.swift into a class that can be usable from objc.
It also gives human readable names to latest iPhones.


PiwikTracker/Device.swift
```diff
@@ -14,20 +16,30 @@ public struct Device {
...
-    public let osVersion: String
+    @objc public let osVersion: String
     
     // The screen size
-    public let screenSize: CGSize
+    @objc public let screenSize: CGSize
     
     // The native screen size
-    public let nativeScreenSize: CGSize?
+    @objc public let nativeScreenSize: CGSize
```
brototyp: You changed the nativeScreenSize to non-optional, but for macOS it is not defined. Correct me if I am wrong on that point.

PiwikTracker/Device.swift
```diff
@@ -69,6 +81,12 @@ extension Device {
         case "iPhone9,2":    return "iPhone 7 Plus (GSM+CDMA)"
         case "iPhone9,3":    return "iPhone 7 (Global)"
         case "iPhone9,4":    return "iPhone 7 Plus (Global)"
+        case "iPhone10,1":   return "iPhone 8 (GSM+CDMA)"
+        case "iPhone10,2":   return "iPhone 8 Plus (GSM+CDMA)"
+        case "iPhone10,3":   return "iPhone X (GSM+CDMA)"
+        case "iPhone10,4":   return "iPhone 8 (Global)"
+        case "iPhone10,5":   return "iPhone 8 Plus (Global)"
+        case "iPhone10,6":   return "iPhone X (Global)"
```
brototyp: Awesome. Thanks!

PiwikTracker/Device.swift
manuroe: My issue was that optional non class parameter like `CGSize?` cannot be represented in objc. I got a build error when prefixing it with `@objc`.
As a Swift rookie, I decided to remove the optionality. Could you advice what is the best thing to do in that situation?

brototyp: Ah, I see your issue there. Yeah. Only pointers to objects can be represented as optionals in Objective-C (null pointer then). 

Since making this a pointer would be very weird I think there would be two solutions. Either define "no native screen size is defined" as `CGSize.zero` or as "is equivalent to screen size". I think the first would be better, since a screen size with no width or height could easily be understood as "not defined". Either way, please add a short comment in the documentation.

brototyp: Lovely. Thanks!

brototyp:
If you merge develop in here I am fine to merge.

manuroe:
> If you merge develop in here I am fine to merge.

@brototyp: done

brototyp:
@manuroe Lovely, thanks! I will add a Changelog entry.

PR #241, opened 2018-04-27
Project Update to Xcode 9.3

ghost:
* updates all Xcode project files to current version 9.3
* adds `.travis.yml` to the Xcode project - for convenience
* removes unnecessary `brew install carthage` config from `.travis.yml`
* makes some changes in class `MemoryQueueSpec` to avoid warnings, local variable `queue` is now a constant


ghost:
@brototyp Please review.

.travis.yml
```diff
@@ -1,12 +1,11 @@
 language: objective-c
+osx_image: xcode9.2
```
brototyp, line 2: We could select xcode9.3 here as well, right?

MatomoTrackerTests/MemoryQueueSpec.swift
```diff
@@ -93,7 +93,7 @@ class MemoryQueueSpec: QuickSpec {
         }
         describe("remove") {
             it("should remove events that are enqueued") {
-                var queue = MemoryQueueFixture.withTwoItems()
+                let queue = MemoryQueueFixture.withTwoItems()
```
brototyp, line 96: 👍

brototyp (approved):
I am happy with this. Feel free to update the osx_image to `xcode9.3` if you feel this to be a worthwhile thing.

.travis.yml
ghost: Yes, but then Swift 4.1 gets selected, which results in deprecation warnings.

brototyp: Alright. Good point. Let's to that when we migrate to 4.1 then. Thanks.

PR #242, opened 2018-04-27
Fixes Xcode settings for Carthage support

### LINKED ISSUE #238: Missing minimum deployment target settings in Xcode project
### What happened
We are using Matomo in a project, which has an iOS and tvOS application. While on iOS everything is fine, I permanently ran into an issue on tvOS when importing the `MatomoTracker` module. Xcode throws an error:

screenshot 2018-04-25 at 16 57 10

### What I've done
Our project needs support for tvOS 10.1. So I tweaked carthage's checkouts by setting the missing deployment target values for tvOS (to 10.0), built the modules again and everything worked as expected.

#### Before Change

beforechange

You can see, the minimum deployment target settings for tvOS, watchOS and macOS are missing. In this case Xcode just uses the latest version for each platform.

#### After Change

afterchange

### Conclusion
This project should be updated **asap** (current situation is a *no-go* for us!) with valid deployment target settings for all platforms (or dedicated targets for each platform). If you agree, I can make the changes and will open a pull request than.

ghost:
Fixes #238

* Adds settings to the Xcode project needed for Carthage support
* Adds the `README.md` to the Xcode project - for convenience
* Adds a note regarding Carthage support to `README.md`

@brototyp  **Discussion**:
* Do we really need iOS 8 support here?
* Minimum deployment target for tvOS is set to 10.0 (Carthage), maybe we should sync the value of the Podfile to the same?


ghost:
@brototyp Please review.

MatomoTracker.xcodeproj/project.pbxproj
```diff
@@ -471,14 +475,17 @@
...
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				MACOSX_DEPLOYMENT_TARGET = 10.12;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				TARGETED_DEVICE_FAMILY = "1,2";
+				TVOS_DEPLOYMENT_TARGET = 10.0;
```
brototyp: According to the [MatomoTracker.podspec](https://github.com/matomo-org/matomo-sdk-ios/blob/develop/MatomoTracker.podspec) we support tvOS >= 9.0. I see no reason why that should be wrong. What do you think about that? Should we change this here to 9.0 and be consistent?

brototyp (commented):
Just a minor thing

MatomoTracker.xcodeproj/project.pbxproj
ghost: Good point, you're right. I will change that on my site to be consistent.

brototyp (approved):

brototyp:
Awesome! Thanks a lot

PR #244, opened 2018-04-30
Fix Xcode Build Settings

ghost:
Updated settings:
* `SUPPORTED_PLATFORMS`
* `TARGETED_DEVICE_FAMILY` (new values `3` and `4` for macOS and watchOS)
* `VALID_ARCHS`

These fixes makes Matomo in Apple TV Simulator available (currently it isn't).


ghost:
@brototyp Please review.

brototyp (approved):
Great. Those are the things you tend to miss if you only use it via Cocoapods.

PR #253, opened 2018-06-14
Replaced generic phone model by device model

Pe-te:
For android devices we see the specific device model in the backend, for iOS devices we only see generic "iPad" or "iPhone".
Here we replace the generic string in the user agent with the real device model (e.g. iPhone8,4),  which lets the backend properly display "iPhone SE".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 5.1.2
+* **bugfix** Replaced generic phone model by device model. [#254](https://github.com/matomo-org/matomo-sdk-ios/pull/253)
+
 ## 5.1.1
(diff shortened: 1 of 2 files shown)


brototyp:
Hi @Pe-te, thanks a lot for this PR. This isn't a method that came in mind, but it's a great idea!
I will have a look and get back to you.

brototyp:
I love it! Looks great! Can you do me a favor and add an entry to the CHANGELOG.md? Feel free to reference yourself there.

brototyp (approved):

brototyp:
Thanks a lot! I will wait for the tests to finish (just to make sure everything works well 😊) and will merge.

Pe-te:
Thx! If it ever causes trouble, maybe there's a cleaner way, where you add the model as extra parameter in the brackets, but this works for us right now.

Issue #255, opened 2018-06-18
Content Impressions / Content Interactions


kanjanaSi:
Do you have any plan to implement Content Impressions and Content Interactions?
Documentation: [https://matomo.org/docs/content-tracking/](url)
API: [https://developer.matomo.org/api-reference/content-tracking](url)

brototyp:
This got implemented and will be part of the next release.

brototyp:
Content tracking just got release as part of the 5.2 release.

PR #256, opened 2018-06-18
track content

kanjanaSi:
API: https://developer.matomo.org/api-reference/tracking-api#optional-content-tracking-info


brototyp:
Hi @kanjanaSi, thanks a million for this pull request. (Just for further reference, this PR is regarding #255).

I will digg into it and get back to you.

brototyp:
Thanks a lot. I would love to split a few things up and clarify the API a bit. According to the [api](https://developer.matomo.org/api-reference/tracking-api#optional-event-trackinghttpspiwikorgdocsevent-tracking-info) there are basically two things to track. Content impressions and content interactions. I would love to see this reflected in the `MatomoTracker.swift` API in form of two separate functions.

What do you think about having these two functions:

```Swift
@objc public func trackContentImpression(name: String, piece: String?, target: String?)
@objc public func trackContentInteraction(name: String, interaction: String, piece: String?, target: String?)
```

An alternative to this would be a single function like this, where if the interaction property is set it is an interaction. If not, it is an impression. 

```Swift
@objc public func trackContent(name: String, interaction: String?, piece: String?, target: String?)
```

With a proper documentation I would be fine with either.

brototyp:
Also: Can you please add sample code in the iOS Example application? Just un-hide the tableview row for content tracking and (re-)implement the ViewController. Also, please make sure to add a line to the `CHANGELOG.md`. Feel free to name yourself there.

kanjanaSi:
@brototyp I changed to two separate functions and added sample code.

brototyp:
@kanjanaSi Awesome. Thanks a lot! That looks really good! Only thing left imho is the documentation of the two functions and in the README.md. I guess both should be updated.

kanjanaSi:
@brototyp  I added trackContentImpression and trackContentInteraction in README.md. ^-^

brototyp:
Thanks a lot for your work! I will merge this as soon as the tests succeed and maybe tweak the README a little afterwards.

brototyp (approved):

PR #273, opened 2018-09-07
Goal tracking

ThomasGoehringer:
Implemented a method for tracking goals with their ID and a revenue connected to that goal.


ThomasGoehringer:
I'm not familiar with the tests but when I run them locally they work with iPhone 6 iOS 9.3

brototyp:
@ThomasGoehringer No worries. That was the 42-false-negative. A little CI bug that is haunting me for a few weeks now. I've only ever seen it on CI and it appears very rarely. I just kicked the build once again. Should be green in no time. 

BTW: Thanks for the PR.

brototyp (approved):

brototyp:
Lovely, Thanks!

PR #274, opened 2018-09-10
Track orders

ThomasGoehringer:
When I migrated to the newest version I noticed order tracking to be missing. So I added the function on my own like in https://matomo.org/docs/ecommerce-analytics/#tracking-ecommerce-orders-items-purchased-required.
I left out the `_ects` field because I didn't need it. Someone might add it later.


MatomoTracker/EventSerializer.swift
```diff
@@ -73,13 +78,21 @@ fileprivate extension Event {
...
+                URLQueryItem(name: "ec_st", value: orderSubTotal != nil ? "\(orderSubTotal!)" : nil),
+                URLQueryItem(name: "ec_tx", value: orderTax != nil ? "\(orderTax!)" : nil),
+                URLQueryItem(name: "ec_sh", value: orderShippingCost != nil ? "\(orderShippingCost!)" : nil),
+                URLQueryItem(name: "ec_dt", value: orderDiscount != nil ? "\(orderDiscount!)" : nil),
                 ].filter { $0.value != nil }
 
             let dimensionItems = dimensions.map { URLQueryItem(name: "dimension\($0.index)", value: $0.value) }
             let customItems = customTrackingParameters.map { return URLQueryItem(name: $0.key, value: $0.value) }
             let customVariableItems = customVariables.count > 0 ? [URLQueryItem(name: "_cvar", value: customVariableParameterValue())] : []
+            let ecommerceOrderItemsAndFlag = orderItems.count > 0 ? [URLQueryItem(name: "ec_items", value: orderItemParameterValue()), URLQueryItem(name: "idgoal", value: "0")] : []
```
brototyp, line 99: Hm, the `idgoal` value is very interesting. I am a little scared that this might (in the future) break the goal feature. I will think about this, if we can solve this any different.

MatomoTracker/MatomoTracker.swift
```diff
@@ -277,6 +277,21 @@ extension MatomoTracker {
...
+    ///
+    /// - Parameters:
+    ///   - id: The unique ID of the order
+    ///   - items: The array of items to be ordered
+    ///   - revenue: The grand total for the order (includes tax, shipping and subtracted discount)
+    ///   - subTotal: The sub total of the order (excludes shipping)
+    ///   - tax: The tax amount of the order
+    ///   - shippingCost: The shipping cost of the order
+    ///   - discount: The discount offered
+    public func trackOrder(id: String, items: [OrderItem], revenue: Float, subTotal: Float? = nil, tax: Float? = nil, shippingCost: Float? = nil, discount: Float? = nil) {
```
brototyp, line 301: Do you have an Idea how to add Obj-c compatibility while keeping nice and clean Swift code?

brototyp:
Thanks a lot for this PR. I guess we should add the `_ects` and add an example to the Example App. Also, there are a few minor questions in the code I am still thinking about.

MatomoTracker/MatomoTracker.swift
ThomasGoehringer: No sorry, I'm not familiar with ObjC

ThomasGoehringer:
I'm gonna have a look into the `_ects` and the Example App.

ThomasGoehringer:
Should `_ects` just be another paramater? I might have a look at it.

MatomoTracker/Event.swift
```diff
@@ -92,10 +92,21 @@ public struct Event {
...
+    /// Ecommerce Order tracking
+    /// https://matomo.org/docs/ecommerce-analytics/#tracking-ecommerce-orders-items-purchased-required
+    let orderId: String?
+    let orderItems: [OrderItem]
+    let orderRevenue: Float?
+    let orderSubTotal: Float?
+    let orderTax: Float?
+    let orderShippingCost: Float?
+    let orderDiscount: Float?
+    let lastOrderTimestamp: Int?
```
brototyp: What do you think about naming this `orderLastTimestamp` to make the naming closer to the oder `order` variables?

brototyp: Also, please change this to be an `NSDate` so it is clearer what exactly this is. When serializing this we can always convert this to a unix timestamp.

MatomoTracker/MatomoTracker.swift
```diff
@@ -287,6 +287,22 @@ extension MatomoTracker {
...
+    ///   - items: The array of items to be ordered
+    ///   - revenue: The grand total for the order (includes tax, shipping and subtracted discount)
+    ///   - subTotal: The sub total of the order (excludes shipping)
+    ///   - tax: The tax amount of the order
+    ///   - shippingCost: The shipping cost of the order
+    ///   - discount: The discount offered
+    ///   - lastOrder: The UNIX timestamp of the customer's last ecommerce order
+    public func trackOrder(id: String, items: [OrderItem], revenue: Float, subTotal: Float? = nil, tax: Float? = nil, shippingCost: Float? = nil, discount: Float? = nil, lastOrder: Int? = nil) {
+        let event = Event(tracker: self, action: [], orderId: id, orderItems: items, orderRevenue: revenue, orderSubTotal: subTotal, orderTax: tax, orderShippingCost: shippingCost, orderDiscount: discount, lastOrderTimestamp: lastOrder)
+        queue(event: event)
```
brototyp, line 305: What do you think about "hiding" the last oder from the user of the SDK? So whenever the user calls `trackOrder` we store the current timestamp and then use that for the next call to `trackOrder`?

brototyp:
> Should _ects just be another paramater? I might have a look at it.

Yeah. Looks great. I added a few comments.

MatomoTracker/MatomoTracker.swift
brototyp: I like your changes. What do you think about moving to the `MatomoUserDefaults`? This way we aren't relying on a singleton but could actually instantiate multiple `MatomoTracker` instances. Additionally we are a bit more consistent with the usage of the UserDefaults in all the other places of the SDK.

README.md
```diff
@@ -147,6 +147,14 @@ The Matomo iOS SDK supports [goal tracking](https://matomo.org/docs/tracking-goa
 matomoTracker.track(goalWithId: 1, revenue: 99.99)
 ```
 
+### Order Tracking
+
+The Matomo iOS SDK supports [order tracking](https://matomo.org/docs/ecommerce-analytics/#tracking-ecommerce-orders-items-purchased-required).
+
+```Swift
+matomoTracker.trackOrder(id: String, items: [OrderItem], revenue: Float, subTotal: Float, tax: Float, shippingCost: Float, discount: Float, lastOrder: Int)
```
brototyp: Please rather try to show the usage of the API.

For example

```Swift
let items = [
   OrderItem(sku: "product_sku_1", name: "iPhone Xs", category: "phone", price: 1086.41, quantity: 1),
   OrderItem(sku: "product_sku_1", name: "iPhone Xs Max", category: "phone", price: 1195.16, quantity: 1)
]

matomoTracker.trackOrder(id: "order_id_123456789", items: items, revenue: 2281.57, subTotal: 2098, tax: 183.57)
```

MatomoTracker/OrderItem.swift
```diff
@@ -0,0 +1,28 @@
...
+    /// The category of the order item
+    let category: String
+    
+    /// The price of the order item
+    let price: Float
+    
+    /// The quantity of the order item
+    let quantity: Int
+    
+    public init(sku: String, name: String, category: String, price: Float, quantity: Int) {
```
brototyp: I just noticed, that only the sku is a required parameter for an item. What do you think about changing the others to be optional?

MatomoTracker/MatomoTracker.swift
brototyp: Alright. I guess we keep it like this for now and consider to add an objc wrapper later.

brototyp (changes requested):

brototyp (approved):

brototyp:
Awesome! Thanks a lot for your work. I will merge this now and it will be included in the next release.

ThomasGoehringer:
No problem :)

PR #275, opened 2018-10-01
update README

kanjanaSi:
When I Added a function to track content, I forgot to remove it from to do list.


brototyp:
Thanks a lot!

brototyp (approved):

Issue #285, opened 2019-02-01
matomo.php vs piwik.php


oskitar89:
In the web site I don't receive any track. Appear this javascript code and it add matomo.php to the home web, but, in the library the website has to end with piwik.php. Any solution?

javascript

<script type="text/javascript">
  var _paq = window._paq || [];
  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://www.example.com/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '32']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>

mobile

@objc convenience public init(siteId: String, baseURL: URL, userAgent: String? = nil) {
        assert(baseURL.absoluteString.hasSuffix("piwik.php"), "The baseURL is expected to end in piwik.php")
        
        let queue = MemoryQueue()
        let dispatcher = URLSessionDispatcher(baseURL: baseURL, userAgent: userAgent)
        self.init(siteId: siteId, queue: queue, dispatcher: dispatcher)
    }

brototyp:
Hi @oskitar89, thanks for creating this issue. Correct me if I am wrong, but as I understood it, you are currently initializing the `MatomoTracker` with a url ending in `matomo.php` but it is failing, because it requires a url ending in `piwik.php`.

If I am correct here, there are two things to do.
1. We should change the check, so that it checks for either of both.
2. In the meantime, you can just change your url in such a way that it ends in `piwik.php`. 

The backend now defaults to `matomo.php` but both work and always will.

oskitar89:
It doesn’t work with matomo.php or piwik.php. We are seeing in the log that all of events are correctly sended. But at the end we are not receveing nothing on the website. What we can do?

brototyp:
Is it possible, that you use a "wrong" filter on the website? If I recall correctly, per default it excludes today, which might lead to what you are seeing as "no events are received".

oskitar89:
We created that site yesterday, so today we have to see the events from yesterday, right?
We have others sites created (all are web and not mobile) and are working correctly. Anything else we can do? And that “filter”, where I can find it? In order to change it?
[...]

PR #286, opened 2019-02-03
Added Support for matomo.php server urls

brototyp:
This implements a fix for not recognized `matomo.php` server urls (#285).


brototyp:
This was released in v6.0.1.

PR #296, opened 2019-05-14
add - add removeAll function

HungryManBB:


brototyp:
Hi @hungryAndCoder, thank you for your PR!

This function sounds like something highly specific. Can you explain in what kind of context you would use this?

MatomoTracker/MatomoTracker.swift
```diff
@@ -399,6 +399,25 @@ extension MatomoTracker {
...
         })
     }
+    
+    /// remove all without ignore
+    @objc public func removeAll(ignoreIndexs indexs: [NSNumber]? = nil) {
+        if let array = indexs, array.count > 0 {
+            
+            let tempArray = dimensions
+            dimensions.removeAll()
+            for i in array {
```
minitrue22, line 409: The time complexity of this implementation is not great, O(n2)
(probably not a big practical problem since the array will not contain that many elements)

MatomoTracker/MatomoTracker.swift
```diff
@@ -399,6 +399,25 @@ extension MatomoTracker {
             dimension.index != index
         })
     }
+    
+    /// remove all without ignore
+    @objc public func removeAll(ignoreIndexs indexs: [NSNumber]? = nil) {
```
minitrue22, line 404: I think methods only should expose Swift types since the framework is written in Swift.
Any bridging from Objc types is the callers responsibility.

````
func removeAll(ignoreIndexes indexes: Set<Int>? = nil)
````

File: MatomoTracker/MatomoTracker.swift [hunk 3]
```diff
@@ -399,6 +399,25 @@ extension MatomoTracker {
...
+                    dimension.index == i.intValue
+                })
+                if resultArray.count > 0, let customDimension = resultArray.first {
+                    dimensions.append(customDimension)
+                }
+            }
+        } else {
+            dimensions.removeAll()
+        }
+    }
```
minitrue22, line 420: One possible implementation would be

````
public func removeAll(ignoreIndexs indexes: Set<Int>? = nil) {

    guard let indexes = indexes else {
        dimensions.removeAll()
        return
    }

    let filtered = dimensions.filter { indexes.contains($0.index) }
    dimensions = filtered

}
````
(might be some typos here, have not compiled it)

File: MatomoTracker/MatomoTracker.swift [hunk 4]
```diff
@@ -399,6 +399,25 @@ extension MatomoTracker {
             dimension.index != index
         })
     }
+    
+    /// remove all without ignore
+    @objc public func removeAll(ignoreIndexs indexs: [NSNumber]? = nil) {
+        if let array = indexs, array.count > 0 {
+            
+            let tempArray = dimensions
+            dimensions.removeAll()
```
minitrue22, line 408: Updating the instance variable `dimensions` is not thread safe.
But that is more a general problem and not unique for this PR.

hungryAndCoder:
> Hi @hungryAndCoder, thank you for your PR!
> 
> This function sounds like something highly specific. Can you explain in what kind of context you would use this?

HungryManBB:
> Hi @hungryAndCoder, thank you for your PR!
> 
> This function sounds like something highly specific. Can you explain in what kind of context you would use this?

sometimes,i need set more than 1 special dimension,it is for my app
anyway,Lack of consideration,close it

PR #300, opened 2019-07-01
Support for swift 5.0 and scope page for Variable.

lpz191:
1. Updated to Swift 5.0.
2. Added the page scope for variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
-
+## 7.0.0
+* **improvement** Added the page scope for variable.[#298]
+    (https://github.com/matomo-org/matomo-sdk-ios/issues/298)
+* **improvement** Updated to Swift 4.3.    
 ## 6.0.1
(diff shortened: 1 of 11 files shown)


lpz191:
Now I have added the explain for my change, please review and merge my codes. Thanks a lot.

brototyp:
Hey @lpz191, thanks for your PR. Greatly appreciated.
Two thought I have towards it. 

Please create a separate PR for every task your are implementing. So in this case one for the update to Swift 5 and one for the support of custom variables in the page scope.

Additionally, I was thinking about your implementation of the support of custom variables in the page scope. There are some downsides with your implementation. Firstly, it's unclear what `isVisitScope` references. Additionally, not it's not possible to set page scope and session scope custom variables for one event. What do you think about having one property of the `Event` as an array of custom variables in the session scope, and one as an array of the view scope?

brototyp:
@lpz191 Thanks for your effort on this PR. Swift 5 support was implemented in #306 in the mean time. I'll close this PR for now.

PR #303, opened 2019-07-23
Specify cookies for authenticated communication with Matomo APIs

krzysztofolszewski86:
We have a case, when we set up our own Matomo API and it has to be behind a corporate proxy. Our mobile app performs authentication, but Matomo communication did not use generated cookies. As a result, no request to the API was possible and no statistics were sent. I've added a way to provide cookies for use in that communication.
While this is a niche case, the modification is small and it may be of use to somebody else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 6.0.2
+* **feature** Specify cookies for authenticated communication with Matomo APIs
+
 ## 6.0.1
 * **improvement** Made copyFromOldSharedInstance available from Objective-C. [#282] (https://github.com/matomo-org/matomo-sdk-ios/issues/282)
 * **improvement** Accepting urls ending in `matomo.php` in addition to `piwik.php` when initializing a new instance. [#286] (https://github.com/matomo-org/matomo-sdk-ios/pull/286)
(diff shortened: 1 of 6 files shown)


brototyp:
Hey @krzysztofolszewski86,

have you tried the `HTTPCookieStorage`? 

So for example like this:

```swift
let cookies = HTTPCookie.cookies(withResponseHeaderFields: responseheaderFields, for: url)
HTTPCookieStorage.shared.setCookies(cookies)
```

If I'm not missing something here, the URLSession in the `URLSessionDispatcher` should use that automatically.

krzysztofolszewski86:
Hi Cornelius,

Sorry for not responding sooner. I was on a short vacation and didn't have
time for this.

I can't say that I've tried this solution, but I have to admit that I'm not
really a proper iOS developer :). I've made this modification because we
have an app written in React Native, and I needed a way to set an
authorization cookie from JS code. If I have some time, I'll try to test
this with our app. Also, I've proposed a similar modification for the
Android library and got some feedback from there as well. I hope to make
both changes approved and then submit a modification for
react-native-matomo. For that, iOS and Android code should be similar
enough to work from a single JS API.

czw., 25 lip 2019 o 18:47 Cornelius Horstmann <notifications@github.com>
napisał(a):

brototyp:
I will close this for now. Feel free to re open it again if needed.

PR #304, opened 2019-07-25
Added `@escaping` attributes to the Queue protocol's functions.

eligutovsky:
This fix allows calling completions of the Queue protocol's functions from an asynchronous scope.
For example, the Queue's functions calling on the Main thread, but processing of arguments required background queue.

I have also fixed a Xcode warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.0.2
+
+* **improvement** Allow to use the Queue's completions in asynchronous scope.
+
 ## 6.0.1
 * **improvement** Made copyFromOldSharedInstance available from Objective-C. [#282] (https://github.com/matomo-org/matomo-sdk-ios/issues/282)
 * **improvement** Accepting urls ending in `matomo.php` in addition to `piwik.php` when initializing a new instance. [#286] (https://github.com/matomo-org/matomo-sdk-ios/pull/286)
--- a/MatomoTracker/EventAPISerializer.swift
+++ b/MatomoTracker/EventAPISerializer.swift
@@ -124,7 +124,7 @@ fileprivate extension DateFormatter {
 fileprivate extension CharacterSet {
     
     /// Returns the character set for characters allowed in a query parameter URL component.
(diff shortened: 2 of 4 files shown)


brototyp:
Hi @eelijah, 

thanks for your PR here. Looks like you implemented your own queue and are actually using this protocol then. Just out of interest: What's your use case and what kind of a queue did you implement?

Regarding the PR itself: Looks great so far. There is one thing I thought about: If we define these closures as `@escaping` we might run into the risk of retain cycles on the call site. For example [here])(https://github.com/matomo-org/matomo-sdk-ios/blob/develop/MatomoTracker/MatomoTracker.swift#L170). If the queue keeps a reference to the closure, the closure to the tracker and the tracker to the queue. Even though this might be quite rare, I guess we should update the callsite as well. What do you think?

eligutovsky:
Hi @brototyp,

Regarding the retain cycle, I agree with you. I added a weak capturing of self to the PR. Does it look like there are no more potential places with a retain cycle with the queue? 

About intent: I'm working on an Analytics SDK that uses Matomo as a component. There are two essential parts:
- First, to use as less as possible the Main thread's resources. 
- To pass all events to the server-side. As a solution used a persistent layer.

``` swift
import MatomoTracker

class MyQueue: Queue {
    private let dispatchQueue = DispatchQueue(label: "sdk.utility.queue")

    var eventCount: Int {
        return dispatchQueue.sync {
            return storage.events.count
        }
    }

    func first(limit: Int, completion: @escaping ([Event]) -> Void) {
        dispatchQueue.async { [storage] in
            completion(storage.events[0 ..< min(limit, storage.events.count)]))
        }
    }

    func remove(events: [Event], completion: @escaping () -> Void) {
        dispatchQueue.async { [storage] in
            let remainingEvents = storage.events - events
            storage.saveEvents(remainingEvents)
            completion()
        }
    }

}
```

brototyp:
Hey @eelijah, thanks for updating the PR. Looks great!

Also thanks for telling me your plans. Sounds great! Looking forward to the SDK.

PR #305, opened 2019-08-02
Fixed crash due to usage of "matomo.php" suffix in the url.

bharath-dev:


brototyp:
Hi @bharath-dev, thanks for your PR. This change was already implemented in #286 and was released in [version 6.0.1](https://github.com/matomo-org/matomo-sdk-ios/releases/tag/v6.0.1).

bharath-dev:
Thank you @brototyp , It was missing when I used the **pod 'MatomoTracker', '~> 6'** as specified in the documentation. Could you please update in documentation as well.

brototyp:
Hey @bharath-dev, `~> 6` is actually not wrong as a version specifier. This includes `6.0.1`. But, to be a bit more specific, what do you think about changing it to `~> 6.0`? That would include everything in the range `>= 6.0` and < `7.0`. What do you think?

PR #314, opened 2019-09-12
Added more escaping symbols; Added test for Event escaping

### LINKED ISSUE #313: Incorrect percent encoding for "+" character
### How to reproduce

Send an event with custom dimension that contains "+" characters in value.

### Expected result

Observe a same value as it was sent from device

### Actual result

Value doesn't contains any + characters.

### Affected version

Reproducible on version 7.0.0

### Possible solution:
https://stackoverflow.com/a/41568692/2798503

Megal:
Closes #313


brototyp:
Hey @Megal, thanks for your PR and sorry it took so long. I tweaked it a tiny bit and merged it here #357

PR #315, opened 2019-09-16
Fix validation of a forcedVisitorId value by using unsigned int type.

eligutovsky:
Found that setting a `forcedVisitorId` with the hex values of more than "7999999999999999" cannot be valid in the current implementation. The solution is just to use `UInt64` instead of `Int`.

Example:
```swift
var a = "7999999999999999"
var b = "8000000000000000"

Int(a, radix: 16) /// Value is 8762203435012037017
UInt64(a, radix: 16) /// Value is 8762203435012037017


Int(b, radix: 16) /// Value is nil
UInt64(b, radix: 16) /// Value is 9223372036854775808
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.0.1
+
+* **bugfix** Fixed an issue with a new `forcedVisitorId` value validation.
+
 ## 7.0.0
 * **important** Dropped support for iOS 8 and 9. [#306](https://github.com/matomo-org/matomo-sdk-ios/pull/306)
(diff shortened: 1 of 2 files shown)


eligutovsky:
@brototyp A half of forsedVisitorIDs does not work properly. Can you release a hotfix for PR? 
In another way, I would fork and specify a git for a Matomo dependency at my SDK. But there is some limitation for a Cocoapod podspec file: 
```
Podspecs cannot specify the source of dependencies. The `:git` option is not supported. `:git` can be used in the Podfile instead to override global dependencies..
```

brototyp:
Oh shoot. Great catch! Do you mind adding a note in the Changelog? I am fine preparing a release during the week.

brototyp:
Cheers!

brototyp:
I am pushing the 7.0.1 release to CocoaPods right now.

PR #319, opened 2019-09-20
Added timeout property to URLSessionDispatcher init method

### LINKED ISSUE #317: Small timeout in URLSession
While testing analytics in bad network conditions with 5 sec timeout, i've found that many requests interrupts with timeout error.

Can you make timeout property public?

MisnikovRoman:
#### Description
Resolves  [#317](https://github.com/matomo-org/matomo-sdk-ios/issues/317)
In bad network conditions with 5 sec timeout many requests interrupts with timeout error.

#### Solution
Add timeout to URLSessionDispatcher init method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 7.0.2 
+* **improvement** Added `timeout` property to URLSessionDispatcher initialization method
+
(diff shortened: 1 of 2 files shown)


brototyp (approved):
Easy and straightforward. Thanks!

Issue #334, opened 2020-04-29
App extension safe API


janhalousek:
Hi, could we toggle the `APPLICATION_EXTENSION_API_ONLY` option on?
We're tracking in app extension and are seeing warnings in XCode.

brototyp:
Hey @janhalousek, thanks for getting in touch. Can you get a little more in detail what you need, what you tried and what the errors/warnings are that you are seeing?

janhalousek:
Hi @brototyp, sure:

I am building an app extension (specifically iMessage extension) that uses matomo tracking. I succeeded linking the framework and everything seems to be working.
but - 
what is just a step from being perfect - when you link a framework that isn't restricted to app-extension-safe-api, XCode gives you a warning:
`ld: warning: linking against a dylib which is not safe for use in application extensions: /Users/[edited]/Library/Developer/Xcode/DerivedData/[edited]/Build/Products/Debug-iphonesimulator/MatomoTracker.framework/MatomoTracker`

The warning isn't specific to Matomo.
The only way to prevent this is to set the `APPLICATION_EXTENSION_API_ONLY` to `YES` (using UI it's a checkbox titled `Allow app extension API only` in target's General tab).
Since Matomo's code seems to be already complying to app extension api (otherwise it wouldn't build having this option toggled on), in my MR i just checked it and it builds without changes.

brototyp:
Thanks for these details. Just one more question. Are you using CocoaPods to install dependencies? If I remember correctly this should be handled by CocoaPods if you add the `APPLICATION_EXTENSION_API_ONLY=YES` to the target that you are adding `Matomo` to in the Podfile. Doesn't it?

janhalousek:
We're using Carthage, which doesn't do this kind of configuration for us..

brototyp:
Oh, gotcha. Yeah, that makes sense. I'll have a look at your pr.

brototyp:
That was easy. I just merged it. Thanks for your contribution!

This will be part of the next release.

PR #335, opened 2020-04-29
Set application extension api only

janhalousek:
Implements #334


brototyp (approved):
Nice and easy. Thanks a lot!

Issue #342, opened 2020-08-06
Outlink tracking


larsriehn:
A direct way to track outlinks is missing in MatomoTracker

brototyp:
That's correct. For now this can only be achieved using [custom events](https://github.com/matomo-org/matomo-sdk-ios#sending-custom-events)

PR #343, opened 2020-08-06
Tracking of outlinks (#342)

larsriehn:


MatomoTracker/MatomoTracker.swift
```diff
@@ -462,3 +462,15 @@ extension MatomoTracker {
...
+
+extension MatomoTracker {
+    /// Track an outlink
+    ///
+    /// - Parameters:
+    ///   - url: The url of the outlink
+    @objc public func trackOutlink(url: URL) {
+        var customTrackingParameters = [String:String]()
+        customTrackingParameters["link"] = url.absoluteString
+        track(Event(tracker: self, action: [], url: url, customTrackingParameters: customTrackingParameters))
```
brototyp, line 474: The `customTrackingParameters` dictionary was meant for actual custom tracking parameters. Since in this case this isn't really a custom tracking parameter if it is defined in the SDK, what do you think about if we move this parameter into the Event itself?

brototyp:
Closed due to lack of activity.

PR #349, opened 2020-09-18
Because the SDK persists this visitor id on app start, then we recommend to ask users for consent before tracking your app users.

mattab:
Adding a quick note to the Readme, after noticing we persist a visitor ID on app starts. So this would mean under GDPR, that we may need to ask app users for consent? I know many people ignore this so far, but it'd be useful to mention it in our guide. feel free to adjust/improve


brototyp:
Hey @mattab, thanks for the PR. Perfectly correct. Let's just merge as it is 😊

Issue #360, opened 2020-11-11
Can I use customTrackingParameters to override the default cdt time format?


Isaacwhyuenac:
Hi, I would like to ask if I can pass a `cdt` in `customTrackingParameters` to override the default `cdt` value (ISO8601) value of the tracker? I would want to override it because I want it to be in unix timestamp (13 digits).

brototyp:
Hi @Isaacwhyuenac, I think that should work.

Out of curiosity: Why do you want to use a unix timestamp instead of the current ISO8601 format?

Isaacwhyuenac:
@brototyp thanks for the reply. But it looks like it is not working because there would have two cdt key in the final url string. One in iso8601 and one in unix time stamp. 

Want this feature because we have some tracking events that fire together in some mini seconds difference. It would help to visualise and order the user’s journey more fully.

brototyp:
Oh, your are correct. Currently, there is no way to replace the value of a parameter. You can only add some.

I think your use case actually is one that the SDK should handle per default. I'll implement a fix for it.

brototyp:
Fixed and merged. This will be in the next release.

Isaacwhyuenac:
Thank you for the reply, but your changes are somehow not what I am expecting, which is to make cdt value unix timestamp (something like 1607875978284).

brototyp:
> Thank you for the reply, but your changes are somehow not what I am expecting, which is to make cdt value unix timestamp (something like 1607875978284).

That's true. The cdt parameter is still a ISO8601 timestamp, but now it does contain milliseconds. Is that sufficient, or is there a reason to have the unix timestamp specifically?

Isaacwhyuenac:
[...]

PR #370, opened 2021-01-03
Custom parameters can override existing parameters

brototyp:
Implements #360


brototyp:
This PR is replaced with #396

PR #384, opened 2021-04-23
Fix copy from UserDefaults method

yalishanda42:
The `MatomoUserDefaults.copy(from:)` method had an unused parameter and was assumed that it will always be `UserDefaults.standard`


brototyp:
Hey @allexks, thanks for this PR.

Great catch! Quite embarrassing 😊

PR #399, opened 2021-10-08
Export startNewSession to objc

b123400:
Hello, I am using this library from Objective-C and find it useful to call `[tracker startNewSession]`.


brototyp:
Hi @b123400, thanks for the tweak. Easy and straightforward. ✌️

PR #415, opened 2022-06-29
Fixes an issue where the revenue of a goal wasn't sent to the server anymore

### LINKED ISSUE #414: Goals revenue value missing
Hello Devs of MatomoTracker,
I'm using the SDK to track many things in the app and it works great, so thanks a lot for this tool.
But since I've increased the sdk version from 7.4.1 to 7.5.1 the tracking of revenue on goal goes broken.
The goals tracking is still working, but without the revenue value.
So I decide to test this case precisely, and downgraded the sdk version back to 7.4.1 and ta-da, it works again...
To be more precisely, the bug seems to be already in the sdk version 7.5, because after increasing the sdk version to the 7.5 my matomo server didn't receive any revenue values anymore, but still receive goals.

I'm using the MatomoTracker method:
public func trackGoal(id goalId: Int?, revenue: Float?)
to track the goals and the revenue value.

I've searched already for changes between the version 7.4.1 and 7.5 but I didn't found any significant changes.
Maybe did you have any idea what goes wrong?

Thanks

brototyp:
Fixes #414

### DIFF
--- a/MatomoTracker/EventAPISerializer.swift
+++ b/MatomoTracker/EventAPISerializer.swift
@@ -91,10 +91,10 @@ fileprivate extension Event {
                 URLQueryItem(name: "c_i", value: contentInteraction),
                 
                 URLQueryItem(name: "idgoal", value: goalId != nil ? "\(goalId!)" : nil),
-                URLQueryItem(name: "revenue", value: revenue != nil ? "\(revenue!)" : nil),
+                // The revenue parameter can be set via the revenue of a goal or the revenue of an order
+                URLQueryItem(name: "revenue", value: (revenue ?? orderRevenue) != nil ? "\((revenue ?? orderRevenue)!)" : nil),
 
                 URLQueryItem(name: "ec_id", value: orderId),
-                URLQueryItem(name: "revenue", value: orderRevenue != nil ? "\(orderRevenue!)" : nil),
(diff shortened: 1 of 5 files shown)


MatomoTracker/EventAPISerializer.swift
```diff
@@ -91,10 +91,10 @@ fileprivate extension Event {
                 URLQueryItem(name: "c_i", value: contentInteraction),
                 
                 URLQueryItem(name: "idgoal", value: goalId != nil ? "\(goalId!)" : nil),
-                URLQueryItem(name: "revenue", value: revenue != nil ? "\(revenue!)" : nil),
+                // The revenue parameter can be set via the revenue of a goal or the revenue of an order
+                URLQueryItem(name: "revenue", value: (revenue ?? orderRevenue) != nil ? "\((revenue ?? orderRevenue)!)" : nil),
 
                 URLQueryItem(name: "ec_id", value: orderId),
-                URLQueryItem(name: "revenue", value: orderRevenue != nil ? "\(orderRevenue!)" : nil),
```
brototyp, line 97: This line here was the culprit. We were always setting the `revenue` parameter again based on the value of an order. In the case of a goal, the order revenue is always `nil` and thus it removes the `revenue` of the order.

Issue #417, opened 2022-08-25
Reset Session and IDs


t-unit:
Currently there seems to be no way to reset session information (`sessionsCount`, `firstVisit`, ...) as well as the visitor id. I would propose some functionality that enables deleting all stored data in the user defaults for a given `MatomoTracker` instance that keeps the instance usable and in correct state afterwards. 

One use case for this is binding the session information to a user and not to the device. E.g. calling reset after user-logout and not keeping the session count etc.

Happy to supply PR with the proposed feature.

SW-Vincent:
Hi, just a question about this topic : what is the default behaviour of a session then ?

brototyp:
This was implemented in #434 and will be in the next release

PR #418, opened 2022-08-29
Add reset method to MatomoTracker

t-unit:
Implements #417

### DIFF
--- a/MatomoTracker/MatomoTracker.swift
+++ b/MatomoTracker/MatomoTracker.swift
@@ -468,3 +468,24 @@ extension MatomoTracker {
     /// The version of the Matomo SDKs
     @objc public static let sdkVersion = "7.3"
 }
+
+extension MatomoTracker {
+    /// Resets all session, visitor and campaign information.
+    ///
+    /// Dispatches events before restting itself.
+    /// After calling this method this instance behaves like the app has been freshly installed.
+    public func reset() {
+        dispatch()
+        
+        matomoUserDefaults.reset()
+        
+        visitor = Visitor.current(in: matomoUserDefaults)
+        session = Session.current(in: matomoUserDefaults)
+        dimensions = []
+        customVariables = []
+        campaignName = nil
+        campaignKeyword = nil
+        
+        startNewSession()
+    }
+}
--- a/MatomoTracker/MatomoUserDefaults.swift
+++ b/MatomoTracker/MatomoUserDefaults.swift
@@ -99,6 +99,18 @@ internal struct MatomoUserDefaults {
             userDefaults.set(newValue, forKey: MatomoUserDefaults.Key.lastOrder)
         }
     }
+    
+    func reset() {
+        userDefaults.removeObject(forKey: Key.totalNumberOfVisits)
+        userDefaults.removeObject(forKey: Key.currentVisitTimestamp)
+        userDefaults.removeObject(forKey: Key.previousVistsTimestamp)
+        userDefaults.removeObject(forKey: Key.firstVistsTimestamp)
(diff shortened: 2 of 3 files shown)


MatomoTracker/MatomoUserDefaults.swift
```diff
@@ -99,6 +99,18 @@ internal struct MatomoUserDefaults {
             userDefaults.set(newValue, forKey: MatomoUserDefaults.Key.lastOrder)
         }
     }
+    
+    func reset() {
```
brototyp, line 103: I was thinking about the term `reset` here. I was wondering, since this is a wrapper of `UserDefaults` if we want to keep the name close to the names of `UserDefaults`. So something like `removeAllObjects` or so. What do you think?

MatomoTracker/MatomoUserDefaults.swift
```diff
@@ -99,6 +99,18 @@ internal struct MatomoUserDefaults {
...
+        userDefaults.removeObject(forKey: Key.totalNumberOfVisits)
+        userDefaults.removeObject(forKey: Key.currentVisitTimestamp)
+        userDefaults.removeObject(forKey: Key.previousVistsTimestamp)
+        userDefaults.removeObject(forKey: Key.firstVistsTimestamp)
+        userDefaults.removeObject(forKey: Key.clientID)
+        userDefaults.removeObject(forKey: Key.forcedVisitorID)
+        userDefaults.removeObject(forKey: Key.visitorUserID)
+        userDefaults.removeObject(forKey: Key.optOut)
+        userDefaults.removeObject(forKey: Key.lastOrder)
```
brototyp, line 112: I could see the risk, once we add a new property that is stored in the user defaults, that we forget to remove it here as well. What do you think about changing the `internal struct Key` to an `enum Key: String, CaseIterable` so we can iterate over all of them here.

MatomoTracker/MatomoTracker.swift
```diff
@@ -468,3 +468,24 @@ extension MatomoTracker {
...
+        visitor = Visitor.current(in: matomoUserDefaults)
+        session = Session.current(in: matomoUserDefaults)
+        dimensions = []
+        customVariables = []
+        campaignName = nil
+        campaignKeyword = nil
+        
+        startNewSession()
+    }
+}
```
brototyp, line 491: With this function we do have the risk, that if we add any properties, we forget to reset them in here as well. Unfortunately I don't have a great idea how to prevent this. I still wanted to bring this up here in case you have an idea.

One option would certainly be to replace the instance of `MatomoTracker`. But since the SDK doesn't provide any instance, this would be left for the user and all we could offer in this function here would be to reset the persisted data. I don't think that's great.

MatomoTracker/MatomoTracker.swift
```diff
@@ -468,3 +468,24 @@ extension MatomoTracker {
     /// The version of the Matomo SDKs
     @objc public static let sdkVersion = "7.3"
 }
+
+extension MatomoTracker {
+    /// Resets all session, visitor and campaign information.
+    ///
+    /// Dispatches events before restting itself.
+    /// After calling this method this instance behaves like the app has been freshly installed.
+    public func reset() {
```
brototyp, line 477: I think, first think in this function, we should ensure we are running on the main thread.

brototyp (commented):
Lovely! Thanks a lot. Looks really good. I really like how you adapted to the coding style of this repo! 🙇 

There are a few thoughts from my side. None of them are really critical but rather thoughts about how to implement.

brototyp:
This PR is replaced by #434

PR #381, opened 2021-03-24
[Devices] added watchOS support

sathoeni:
Added support for watchOS.
Successfully tested and integrated in a watchOS Standalone App

PR #445, opened 2024-02-27
[Devices] added watchOS support (updated develop branch)

dderg:
This is the same as #381, but with latest develop merged
Tested that it actually works 
image


dderg:
I think the danger check is broken, or please let me know what can I do to make it pass, I think it's complaining that parent commit is coming from a fork branch

brototyp:
Hi @dderg, thanks a lot for taking care and for your patience with me for this one. The danger check is not on you, but on me. I'll need to look into that one. I'll merge away.

brototyp:
Merged and part of Version 7.6.0

PR #448, opened 2024-03-22
Privacy Manifest

felipe-nipper:
- Added Privacy Manifest file as required by [Apple](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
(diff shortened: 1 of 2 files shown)


brototyp:
Thanks for the contribution! LGTM. I'll let the tests pass and merge

brototyp:
Merged and part of Version 7.6.0

pau-lu:
Hi @felipe-nipper -  a quick question - does the manifest only disclose NSPrivacyAccessedAPITypes because NS PrivacyTracking, NSPrivacyTrackingDomains and NSPrivacyCollectedDataTypes are not applicable? The 2nd note here https://developer.apple.com/documentation/bundleresources/privacy_manifest_filesnote also says "You only need to supply NSPrivacyAccessedAPITypes for apps and 3rd party SDKs on iOS" but I think this means that NSPrivacyAccessedAPITypes dictionary is required in those SDKs in addition to the 3 above (to the extent they are applicable). Thank you.

felipe-nipper:
Hi @pau-lu, the values of NSPrivacyTracking, NSPrivacyTrackingDomains, and NSPrivacyCollectedDataTypes can change based on how you use the data. For instance, your usage of Matomo doesn't necessarily fit into the tracking requirements set by Apple. Hence, the values you're required to delineate can vary by project. Keeping this in mind, you should input this information at your app level and not in the third-party SDK. Additionally, it seems that NSPrivacyTrackingDomains and NSPrivacyCollectedDataTypes are not necessarily required.

pau-lu:
Hi @felipe-nipper - thank you!

PR #455, opened 2024-08-13
Make CustomDimension fields public

idrougge:
`CustomDimension` is a public struct, but its fields aren't, which makes adding (retroactive) conformance to e.g. `CustomStringConvertible` and `CustomDebugStringConvertible` impossible.


brototyp:
Thanks a lot. Straight forward

PR #464, opened 2025-03-04
Conform spelling of `ID` to Swift API design guidelines

idrougge:
https://www.swift.org/documentation/api-design-guidelines/#general-conventions states that
> Acronyms and initialisms that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions:
> ```
> var utf8Bytes: [UTF8.CodeUnit]
> var isRepresentableAsASCII = true
> var userSMTPServer: SecureSMTPServer 
> ```

This includes the initialism "ID"; one might note that it is uniformly spelt with either all uppercase or all lowercase in all Swift standard library and Apple APIs.

This PR would align Matomo, which currently uses a mix of `ID` and `Id`, with those guidelines. Fallbacks with fixits have been provided for all public symbols.


MatomoTracker/Event.swift
```diff
@@ -138,4 +138,9 @@ extension Event {
         self.orderLastDate = orderLastDate
         self.isCustomAction = isCustomAction
     }
+
+    @available(*, unavailable, renamed: "init(tracker:action:url:referer:eventCategory:eventAction:eventName:eventValue:customTrackingParameters:searchQuery:searchCategory:searchResultsCount:dimensions:variables:contentName:contentInteraction:contentPiece:contentTarget:goalID:revenue:orderID:orderItems:orderRevenue:orderSubTotal:orderTax:orderShippingCost:orderDiscount:orderLastDate:isCustomAction:)")
```
brototyp: Is there a reason why you chose `unavailable` here vs `deprecated`?

brototyp (approved):
Hey @idrougge, thanks for the PR. 
Looks good to me and I am happy to merge it though. Just one small question.

MatomoTracker/Event.swift
idrougge: Sorry, no, that was a leftover from work in progress. Fixed now.

PR #470, opened 2025-11-14
remove url from event encoding

rodolphebossin:
In EventAPISerializer, all query parameters—including url—are percent-encoded.
However, the url parameter passed to Event is already percent-encoded and gets encoded twice.

For example, Cancérologie is already encoded to Canc%C3%A9rologie.
But then it is encoded again and becomes Canc%25C3%25A9rologie

This creates a mismatch between platforms:
	•	Android sends the correctly encoded URL (single encoding)
	•	iOS sends the double-encoded version

As a result, Matomo displays two separate entries for what should be the same screen/page.

Android:
/Accueil/Cancérologie

iOS: 
/Accueil/Canc%C3%A9rologie

This leads to duplicated pages in the Matomo Dashboard:

Android : 
Capture d’écran 2025-11-14 à 10 31 45

iOS : 
Capture d’écran 2025-11-14 à 10 32 01

This change aligns the iOS SDK with Android behavior, avoids dashboard duplication, and ensures consistent analytics when URLs contain non-ASCII characters.

### DIFF
--- a/MatomoTracker/EventAPISerializer.swift
+++ b/MatomoTracker/EventAPISerializer.swift
@@ -2,11 +2,14 @@ import Foundation
 
 final class EventAPISerializer {
     internal func queryItems(for event: Event) -> [String: String] {
-        event.queryItems.reduce(into: [String:String]()) {
+        var queryItemsEncoded = event.queryItems.reduce(into: [String:String]()) {
             $0[$1.name] = $1.value
         }.compactMapValues {
             $0.addingPercentEncoding(withAllowedCharacters: .urlQueryParameterAllowed)
         }
+        queryItemsEncoded["url"] = event.url?.absoluteString // already percent encoded
+        return queryItemsEncoded
+
     }
     
     internal func jsonData(for events: [Event]) throws -> Data {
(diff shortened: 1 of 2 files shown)


brototyp:
Hi @rodolphebossin, thanks a lot for the report and the PR. Highly appreciated. I understand that it's not great if both track in different ways. I looked into it and can't see a functional difference between the two. 

If I track a url like so ...
```MatomoTracker.shared.track(eventWithCategory: "TestCategory", action: "TestAction", name: "TestName", value: 7, url: URL(string: "http://fr.wikipedia.org/wiki/Été"))```

... it does look encoded on the backend, but it functions fine. One can click the url and it functions correctly.
Bildschirmfoto 2025-11-22 um 12 37 41

I am wondering what the "correct" behavior is. @mattab do you have any input in this regard?

rodolphebossin:
Hi @brototyp ,
you're right, it does function, but if you do the same test on the Android side, you'll get a tracking reference (non encoded).
The tracking does work but creates different records for Android and iOS when it should only create one split by system.

mattab:
@brototyp Maybe this particular URL is double-encoded before sent to Matomo? 

But I agree with the bug report author, that the URL should display correctly in Matomo (without double encoding) so it indicates an issue in the tracking (somewhere) 👍

brototyp:
I digged a bit more into this. Indeed, the URL was encoded twice. I fixed the issue in a different way though. This bug will be fixed in the next release.

